PR #3, opened 2022-06-03
Swap to the proper package path.

morrowc:
I noticed, while working through the accounting proto,
that the package path was incorrect in this proto.


aashaikh (approved):
LGTM - thanks for the fix

PR #21, opened 2022-07-13
Update README.md

nmahabaleshwar:
Updated the example to include the source

### DIFF
--- a/authz/README.md
+++ b/authz/README.md
@@ -161,10 +161,12 @@ will be able to call any of the `gNSI.ssh` RPCs.
   "name": "gNSI.ssh policy",
   "allow_rules": [{
     "name": "admin-access",
-    "principals": [
-      "spiffe://company.com/sa/alice",
-      "spiffe://company.com/sa/bob"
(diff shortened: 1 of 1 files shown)


morrowc (commented):
$ gpg --search 4AEE18F83AFDEB23
gpg: data source: https://keys.openpgp.org:443
(1)	  2048 bit RSA key 4AEE18F83AFDEB23, created: 2017-08-16
Keys 1-1 of 1 for "4AEE18F83AFDEB23".  Enter number(s), N)ext, or Q)uit > 1
gpg: key 4AEE18F83AFDEB23: new key but contains no user ID - skipped
gpg: Total number processed: 1
gpg:           w/o user IDs: 1


Uhm, a key without a userid or identity is as good as no key at all... Wanna make a real key and perhaps use that? (I guess who knows if you really are who you say you are... without some signatures on the key from other known/trustworthy signers though...)

morrowc:
this seems to have gotten fixed in the background, closing pr.

PR #22, opened 2022-07-13
Make the FinalizeResponse and UploadResponse format the same.

morrowc:
I think the proper proto style is the UploadResponse.


morrowc:
Forgot to set marcus/tomek as reviewers (Ideally that's done now? github != work workflow)

tomek-US (approved):
LGTM

PR #25, opened 2022-07-14
Initial take on the index.md for cert/

morrowc:
This should sketch out the reason for being along
with some use-cases for the rpcs. This needs
at least 1 example of updating more than one of the PKI elements.

### DIFF
--- a/cert/index.md
+++ b/cert/index.md
@@ -1,0 +1,103 @@
+# gNSI cert Service Protobuf Definition
+**Contributors**: hines@google.com, morrowc@google.com, tmadejski@google.com
+**Last Updated**: 2022-07-12
+
+## Background
+
+The cert service definition provides the API to be used for installing,
+rotating, and testing PKI primatives used on network systems. The Rotate
+and Install are bidirectional streaming rpcs which permit mutating
+Certificates, Root Certificate Bundles, Certificate Revocation Lists.
+For either an Install or Rotate stream it is possible to mutate one or
+more of the elements, and to send a Finalize message once the in-flight
+change has been verified to be operational. Failure to send the Finalize
+message will result in the candidate element being discarded and the
+original element being used instead.
+
+## Motivation
+
+Management of the PKI elements for a network system should have
+a clear and direct method for installation and update.
+
+### CertificateManagement.Install
+
+CertificateManagemetn.Install will permit installation, and
+verification of function, of any of the PKI elements. The normal
+use-case would be to:
+
+* send an Certificate and Key to a network system as an
+InstallCertificateRequest.
+* verify that the services which will use the certificate continues
+to operate normally.
+* send a FinalizeRequest to finish the installation process.
+
+### CertificateManagement.Rotate
+
+CertificateManagement.Rotate will permit rotation, and
+verification of function, of any of the PKI elements.
+The normal use-case would be to:
+
+* send an CertificateBundle to a network system as a
+RotateCertificateRequest.
+* verify that the services which will use the new certificate bundle
+continue to operate normally.
+* send a FinalizeRequest to finish the rotateion process.
+
+## User Experiences
+
+### A Certificate is to be installed
+
+Create a new Certificate and Key.
+
+Send that certificate to the target network system with a
+cert.InstallCertificateRequest to the cert.Install rpc. The
+InstallCertificateRequest's install_request will be a
+cert.Certificate.
+
+Verify that the certificate newly deployed is usable by the relevant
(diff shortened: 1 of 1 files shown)


cert/index.md
```diff
@@ -0,0 +1,103 @@
+# gNSI cert Service Protobuf Definition
+**Contributors**: hines@google.com, morrowc@google.com, tmadejski@google.com
+**Last Updated**: 2022-07-12
+
+## Background
```
danielywong, line 5: Would this be the right place to also introduce the concept behind Entity ID that allows the gNSI server to manage more than one set of credentials on the host?  Most will think of X.509 being issued to a host and might therefore get confused when first looking into the example flows in the .proto (note labels "1A", "1B", ...etc).

cert/index.md
```diff
@@ -0,0 +1,103 @@
...
+verification of function, of any of the PKI elements.
+The normal use-case would be to:
+
+* send an CertificateBundle to a network system as a
+RotateCertificateRequest.
+* verify that the services which will use the new certificate bundle
+continue to operate normally.
+* send a FinalizeRequest to finish the rotateion process.
+
+## User Experiences
```
danielywong, line 46: Suggest "User Journey"

cert/index.md
```diff
@@ -0,0 +1,103 @@
...
+### A CertificateRevocationList is to be installed
+
+Create a new CertificateRevocationList (CRL).
+
+Send that certificate to the target network system with a
+cert.InstallCertificateRequest to the cert.Install rpc. The
+InstallCertificateRequest's install_request will be a
+cert.CertificateRevocationListBundle.
+
+Verify that the CRL newly deployed is usable by the relevant
```
danielywong, line 76: "Optionally verify"?

cert/index.md
morrowc: If you have words here that'd be fine by me.
I banged away 'tousand monkeys' style in order to get this far :)

cert/index.md
morrowc: This is (the header) just copy/pasted from the healthz example, I don't have a preference, but I expect consistency matters across the docs of this type.

cert/index.md
morrowc: this, yes :)

cert/index.md
```diff
@@ -0,0 +1,103 @@
+# gNSI cert Service Protobuf Definition
+**Contributors**: hines@google.com, morrowc@google.com, tmadejski@google.com
+**Last Updated**: 2022-07-12
+
+## Background
+
+The cert service definition provides the API to be used for installing,
+rotating, and testing PKI primatives used on network systems. The Rotate
```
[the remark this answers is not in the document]
morrowc: I think we didn't agree on how to delete, we should definitely have an answer though :)

For now I'll add , delete, ....  we can sort out HOW that works later.

cert/index.md
```diff
@@ -0,0 +1,103 @@
...
+and Install are bidirectional streaming rpcs which permit mutating
+Certificates, Root Certificate Bundles, Certificate Revocation Lists.
+For either an Install or Rotate stream it is possible to mutate one or
+more of the elements, and to send a Finalize message once the in-flight
+change has been verified to be operational. Failure to send the Finalize
+message will result in the candidate element being discarded and the
+original element being used instead.
```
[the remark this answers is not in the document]
morrowc: I think that's the last sentence:
  "Failure to send the Finalize message will result in the candidate element being discarded and the
original element being used instead."

cert/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+Verify that the CertificateBundle newly rotated is used by services
+which require it.
+
+Send a Finalize message to the cert.Rotate rpc to close out the action.
+
+If the stream is disconnected prior to the Finalize message being
+sent, the proposed configuration is rolled back automatically.
+
+## Open Questions/Considerations
```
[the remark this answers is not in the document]
morrowc: I added a rollbck example. (session breaks)

today 'UploadResponse' has no error return.
We probably need to add an error string in this response. (if we expect to send back errors)

cert/index.md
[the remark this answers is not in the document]
marcushines: so some of the clarity goes on both sides - if the server crashes or someting interupts the stream - we need to make sure it is clear to service implementor that they must handle this rollback properly.
like in service crash... how is the transaction rolled back in service restart

marcushines (approved):

PR #27, opened 2022-07-15
Removing unbounded collection of pathz policies. Now active + sandbox pathz policies only.

tomek-US:
* Removes Install()
* Adds Get()
* Adds PolicyInstance enum and adds its use to Probe() and Get()
* Removes 'Pathz' from RotatePathz{Request|Response}

### DIFF
--- a/pathz/pathz.proto
+++ b/pathz/pathz.proto
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 Google Inc. All Rights Reserved.
+// Copyright 2021, 2022 Google Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -63,15 +63,18 @@ service Pathz {
   //     Client <-- UploadResponse <--- Target
   //
   //   Step 3 (optional): Test/Validation by the client.
-  //     During this step client attempts to call a RPC that is allowed
-  //     in the new policy and validates that the new policy "works".
-  //     Additionally the client should access an OpenConfig path that is
-  //     not allowed and the attempt must fail proving that the OpenConfig
-  //     gNMI Path-based Authorization Policy "works".
+  //     During this step client calls (possibly multiple times)
+  //     the Probe(POLICY_INSTANCE_SANDBOX) RPC to check that a combination of
+  //     user ID, gNMI path and operation mode that is supposed to be allowed by
+  //     the new policy and validates that the response is ACTION_PERMIT.
+  //     Additionally the client should call the Probe(POLICY_INSTANCE_SANDBOX)
+  //     RPC with a combination of user ID, gNMI path and operation mode that is
+  //     not allowed in the new policy and the attempt must result in
+  //     the ACTION_DENY response proving that the new policy "works".
   //     Once verified, the client then proceeds to finalize the rotation.
-  //     If the verification did not succeed the client will cancel the
+  //     If the verification did not succeed the client will cancel the Rotate
   //     RPC thereby forcing the target to perform a rollback of the new
-  //     OpenConfig gNMI Path-based Authorization Policy.
+  //     OpenConfig gNMI Path-based Authorization Policy to the previous one.
   //
   //   Step 4: Final commit.
   //     Client ---> FinalizeRequest ----> Target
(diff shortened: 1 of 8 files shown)


pathz/pathz.proto
```diff
@@ -72,60 +67,32 @@ service Pathz {
   //     in the new policy and validates that the new policy "works".
```
morrowc, line 67: i think step 3 should be sending 'Probe()' messages and getting replies, not necessarily testing the policy.
The policy may not be 'live' right?

pathz/pathz.proto
```diff
@@ -72,60 +67,32 @@ service Pathz {
...
-  //     If the new verification did not succeed the client will cancel the
+  //     gNMI Path-based Authorization Policy "works".
+  //     Once verified, the client then proceeds to finalize the rotation.
+  //     If the verification did not succeed the client will cancel the
   //     RPC thereby forcing the target to perform a rollback of the new
-  //     OpenConfig Path-based Authz Policy.
+  //     OpenConfig gNMI Path-based Authorization Policy.
   //
   //   Step 4: Final commit.
   //     Client ---> FinalizeRequest ----> Target
```
morrowc, line 77: there isn't  a defined finalizeResponse, what happens if there's an error finalizing?
Is there a reason to not have FinalizeResponse{ string error }  or similar sort of construction?

pathz/pathz.proto
```diff
@@ -72,60 +67,32 @@ service Pathz {
...
   // operation performed by a user on a single gNMI path.
-  // The response is based on the currently active policy and is evaluated
-  // without actually performing the gNMI operation.
+  // The response is based on the instance of policy specified in the request
+  // and is evaluated without actually performing the gNMI operation.
   rpc Probe(ProbeRequest) returns (ProbeResponse);
+
+  // Get returns specified instance of the OpenConfig gNMI Path-based
+  // Authorization Policy together with its version and created-on information.
+  rpc Get(GetRequest) returns (GetRequest);
```
morrowc, line 90: Get(GetRequest) returns (GetResponse)

(you doubled the requests)

morrowc (approved):

pathz/pathz.proto
```diff
@@ -142,49 +109,28 @@ message RotatePathzRequest {
 }
 
 // Response messages from the target.
-message RotatePathzResponse {
+message RotateResponse {
```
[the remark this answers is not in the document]
morrowc: Probably a style thing, on my part or .. 'chris just doesn't know' thing, but...
Why not have a response like this be able to send back an error?
The rotate request MAY have an invalid proto in it, right? or there could be problems on the receiver side with 'using' the incoming data (system disk full, no processing time/deadline, etc).

If it's easy to add later, fine, but I would expect we'd definitely need some error condition handling/replying eventually.
troubleshooting a system that just eats packets ... is hard :( if there's no error channel back.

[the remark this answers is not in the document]
morrowc: 'i was holding it wrong' - ok :) (marcus explained offline)
thnx!

pathz/pathz.proto
tomek-US: You are right - now the policy is in a sandbox until the Finalize message, so there is no other way to check it out than the Probe() RPC.

pathz/pathz.proto
tomek-US: :-) 
Changed

pathz/pathz.proto
```diff
@@ -205,27 +155,89 @@ message UploadRequest {
...
+enum PolicyInstance {
+  // Invalid instance. Referring to this instance in any of the RPCs always
+  // results in an error report.
+  POLICY_INSTANCE_UNSPECIFIED = 0;
+
+  // The policy that is currently used by the gNMI service to authorize access.
+  POLICY_INSTANCE_ACTIVE = 1;
+
+  // The most recent policy that has been uploaded during the Rotation() RPC.
+  // If there is no Rotate() RPC in progress, then referring to this instance
```
danielywong, line 176: Is the concept of "on Rotate() RPC in progress" meant to be per server?  ie, to support one client performs a rotation and another client tries to probe at the same time.  Otherwise, use of Sandbox can be implicitly assumed if the client had issued any rotate (which I believe is the theme in the other gNSI RPC that support test before commit).

marcushines (approved):

pathz/pathz.proto
danielywong: I take what you said as each stream has its own independent sandbox then.  I wonder if that is worth being specific in the comment?  ...just in case someone tries to write a client that use one stream for rotation and a separate one for probing.

danielywong (approved):

PR #32, opened 2022-07-29
Limits number of authz polices to one per service and adds more info.

tomek-US:
- removes the Install() RPC
- replaces policy ID with the name of the service the new policy will be applied to.
- adds Probe() RPC
- adds Get() RPC
- expands the README.md

### DIFF
--- a/authz/README.md
+++ b/authz/README.md
@@ -198,13 +198,13 @@ Authorization Policy, namely:
 1. Starting the `gNSI.authz.Rotate()` streaming RPC.
 
    As the result a streaming connection is created between the server (the
-   switch) and the client (the management aplication) that is used in the
+   switch) and the client (the management application) that is used in the
    following steps.
 
    > **⚠ Warning**
    > Only one `gNSI.authz.Rotate()` can be in progress.
 
-1. The client uplads new gRPC-level Authorization Policy using
+1. The client uploads new gRPC-level Authorization Policy using
    the `UploadRequest` message.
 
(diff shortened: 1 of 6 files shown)


authz/authz.proto
```diff
@@ -204,11 +172,82 @@ message UploadRequest {
...
+  string server = 4;
 }
 
-message UploadResponse {}
+message UploadResponse {
+}
 
+// ProbeRequest contains a single user name, the gRPC service name and the gRPC
+// call being attempted.
+message ProbeRequest {
```
marcushines, line 190: do we think probe in this context is implementable?

marcushines (commented):
everything else LGTM

authz/README.md
```diff
@@ -174,29 +174,152 @@ will be able to call any of the `gNSI.ssh` RPCs.
...
+Authorization Policy, namely:
+
+1. Starting the `gNSI.authz.Rotate()` streaming RPC.
+
+   As the result a streaming connection is created between the server (the
+   switch) and the client (the management application) that is used in the
+   following steps.
+
+   > **⚠ Warning**
+   > Only one `gNSI.authz.Rotate()` can be in progress.
```
marcushines, line 205: the auth policy is "declarative" for all servers and services on the device
All policies must be defined as previous policies will be replaced during the rotate

tomek-US: done.

marcushines (approved):

PR #36, opened 2022-09-13
adding AuthorizedPrincipalsCommand and fixing bundle references

sourcequench:
My comment in credentialz.proto is really a discussion point. The nomenclature "authorized_users" is a Google internal thing - and probably we should also change it to be what sshd_config calls it, AuthorizedPrincipalsFile. This authorized_users file is a per-system-user setting. sshd could be configured in many ways with AuthorizedPrincipalsFile (see man sshd_config) including being either a relative path to the users home dir, or an absolute path for a singular file, and can have tokens passed. In other words, it *could* be a user specific list of authorized principals... or not.

In order to specify AuthorizedPrincipalsCommand, it would *not* be a user-specific setting, but rather a singular command (and arguments, which will pass along user details) represented in the proto changes in this pr by a string.


tomek-US:
Naming things is the most difficult task in computer science :-) therefore there is usually a space for improvement and all comments/suggestions how to make them better are more than welcome.
However, the API is not designed to be OpenSSH-specific. It is designed to work with higher-level concepts like authorized user and not implementation-specific commands to be executed to generate a configuration file and therefore allow other implementations 'under the hood'.

sourcequench:
Agreed, and I'll set whatever name you wish - you tell me the name.
"authorized users" is Google-specific naming for a file that happens to be
used with an OpenSSH-specific feature AuthorizedPrincipalsFile - so we're a
couple layers deep into ssh assumptions there. The RFC based certificate
authentication (as opposed to openssh-specific) uses x509 and is used
roughly nowhere, whereas OpenSSH is widely used across many operating
systems. We wouldn't want to limit ourselves to the venn diagram feature
overlap between wolfSSH, CrushFTP Server ssh... and OpenSSH given that
OpenSSH has a near 100% share and every network operating system I'm aware
of is using it.

credentialz/credentialz.proto
```diff
@@ -442,12 +443,30 @@ message UserPolicy {
     repeated Option options = 2;
   }
 
+  message AuthorizedPrincipalCheck {
+    enum PrincipalCheckTool {
```
marcushines, line 447: enum Tool {
  TOOL_UNSPECIFIED =  0;
  TOOL_HIBA = 1; 
}

credentialz/credentialz.proto
```diff
@@ -442,12 +443,30 @@ message UserPolicy {
...
   // The system role account name (e.g. root). This account must exist.
   string account = 1;
-
-  // List of authorized usernames for this system role as found in the SSH
-  // certificate.
-  repeated SshAuthorizedUser authorized_users = 2;
+  // How the system authorizes users, either by way of a hard-coded repeated
+  // SshAuthorizedUser (authorized_users) or through an AuthorizedUserTool enum.
+
+  // Authroizing users is either through a user keyed explicit list, or can be
```
marcushines, line 460: Authorizing

credentialz/credentialz.proto
tomek-US: +1 (not that we have a choice ;-) )

credentialz/gnsi-credentialz.yang
```diff
@@ -48,18 +48,56 @@ module gnsi-credentialz {
...
+                "The version of the Certificate Authority keys.";
         }
 
-        leaf active-trust-bundle-created-on {
+        leaf active-trusted-user-ca-keys-created-on {
             type created-on;
             description
-                "The timestamp of the moment when the bundle of the Certificate
-                Authority certificates (a.k.a. trust bundle) was created.";
+                "The timestamp of the moment when the trusted user ca keys
```
tomek-US, line 60: nit: CA instead ca?

credentialz/gnsi-credentialz.yang
```diff
@@ -48,18 +48,56 @@ module gnsi-credentialz {
...
+        }
+
+        leaf active-host-certificate-created-on {
+            type created-on;
+            description
+                "The timestamp of the moment when the host certificate
+                 was created.";
+        }
+
+        leaf active-host-certificate-file-version {
```
tomek-US, line 77: What is the difference between this leaf and the one in line 64?

credentialz/gnsi-credentialz.yang
```diff
@@ -48,18 +48,56 @@ module gnsi-credentialz {
...
+            type version;
+            description
+                "The version of the host ssh certificate.";
+        }
+
+        leaf active-host-certificate-file-version-created-on {
```
tomek-US, line 83: Honestly I do not like the idea of 'file' being part of the leaf name. Yes, the key/certificate is in a file (most likely) but what counts is the fact that the key/certificate was created at some point and has the metadata 'version' associated with it. 

Could we get rid of it?

credentialz/gnsi-credentialz.yang
```diff
@@ -48,18 +48,56 @@ module gnsi-credentialz {
         description
             "SSH server credentials freshness-related data.";
 
-        leaf active-trust-bundle-version {
+        leaf active-trusted-user-ca-keys-version {
             type version;
             description
-                "The version of the bundle of the Certificate Authority
-                certificates a.k.a. trust bundle.";
+                "The version of the Certificate Authority keys.";
```
tomek-US, line 54: A grammar question from a not-native speaker: does 'the Certificate Authority keys' cover the case of multiple CAs? Or just multiple keys from one CA?

credentialz/gnsi-credentialz.yang
```diff
@@ -48,18 +48,56 @@ module gnsi-credentialz {
...
+            type created-on;
+            description
+                "The timestamp of the moment when the host certificate
+                 was created.";
+        }
+
+        leaf active-host-certificate-file-version {
+            type version;
+            description
+                "The version of the host ssh certificate.";
```
tomek-US, line 80: nit: SSH instead ssh?

tomek-US (commented):
Thx a lot for spending time to prepare the change request. Appreciate it.

credentialz/credentialz.proto
sourcequench: Done.

credentialz/credentialz.proto
```diff
@@ -442,12 +443,30 @@ message UserPolicy {
...
+  // dynamically determined using a PrincipalCheckTool, perhaps better known as
+  // AuthorizedPrincipalCommand.
+  oneof user_authorization {
+    // List of authorized usernames for this system role as found in the SSH
+    // certificate.
+    repeated SshAuthorizedUser authorized_users = 2;
```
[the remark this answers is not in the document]
sourcequench: Let me know if wrapping these up into a new SshAuthorizedUser**s** message then making that required works out - I think this works?

credentialz/gnsi-credentialz.yang
sourcequench: Ha, from a native speaker I'm not even sure. I think either "Certificate Authority keys" or the possessive "Certificate Authority's keys" would be acceptable (but the latter sounds bad to me). "keys" clearly implies multiple are possible, but yes the singular "Certificate Authority" seems like it could be interpreted as being only one - although it sort of doesn't make sense to have multiple keys for one CA as literally the definition of a CA for openssh is just a keypair that signs stuff - much less complicated than x509.

I have no better name I can think of offhand.

credentialz/gnsi-credentialz.yang
sourcequench: Done.

credentialz/gnsi-credentialz.yang
sourcequench: Yeah, I think this is a mistake. We want to _not_ mention the word file, so yanking this one out.

credentialz/gnsi-credentialz.yang
sourcequench: Agreed. took out file, PTAL

marcushines (approved):

marcushines (approved):

tomek-US (approved):
Thx!

credentialz/credentialz.proto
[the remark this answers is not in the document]
tomek-US: LGTM

PR #37, opened 2022-09-15
fix compilation errors and add _grpc.pb.go files

brianneville:
Adds a compile_protos.sh script to compile .proto files


tomek-US:
Hmm. I am not sure why you need this change. The existing files include code for both the client and the server.

tomek-US:
I have borrowed you fixes to `acct.proto`, added `BUILD.bazel` and generated `acct.pb.go`. Please take a look at #38

brianneville:
> Hmm. I am not sure why you need this change. The existing files include code for both the client and the server.

The main distinction here is that with this change the server interfaces will require the `Unimplemented<ServiceName>Server` to be embedded, with the inclusion of the `mustEmbedUnimplemented<ServiceName>Server()` functions. 
e.g. PathzServer becomes: 
```go
type PathzServer interface {
	Rotate(Pathz_RotateServer) error
	Probe(context.Context, *ProbeRequest) (*ProbeResponse, error)
	Get(context.Context, *GetRequest) (*GetResponse, error)
	mustEmbedUnimplementedPathzServer()
}
```
This allows for implementations to be forward compatible by default (and developers can still always opt-out by making use of the `Unsafe<ServiceName>Server` type).
All of the services in gNOI also generate _grpc.pb.go files in a similar fashion (see [compile_protos.sh](https://github.com/openconfig/gnoi/blob/93cdd9ae9f35d8b4bc1599d0a727b294faeca352/compile_protos.sh)), and it would be useful here in gNSI as well.

Other than that the only difference is that comments and licenses were generated in the .pb.go files, which are always just nice to have.

tomek-US:
I understand but on the other hand those files are those that are generated by the `bazel` `BUILD` files so, by keeping them we get the same behavior/code in both cases: (1) accessing them directly from the repo and (2) generating them using `bazel`.

marcushines:
I think this is resolved now via latest commit

marcushines (approved):

Issue #45, opened 2022-10-19
A more flexible abstraction for gNSI Certz 


brianneville:
Currently, gNSI certz assumes the following:
```
Target (as seen from gNSI.certificate microservice point of view)
 |
 +-+ certificate (used by all gRPC services)
 | +- certificate (with public key)
 | +- private key
 |
 +-+ trust bundle (Certificate Authority certificates)
 | +- CA Root certificate
 | +- CA Intermediate Certificate
 |
 +-+ Certificate Revocation List bundle
   +- Certificate Revocation List A
   +- Certificate Revocation List B

 The above shown topology implies that every operation performed using this
 service changes a credential for all gRPC services at the same time.
 ```

I would like to propose a different abstraction here, based on the idea of an SSL profile which encapsulates all of the certs/bundle/CRLs. 

An SSL profile would have the same structure as above: 
```
SSL profile
 |
 +-+ certificate
 | +- certificate (with public key)
 | +- private key
 |
 +-+ trust bundle (Certificate Authority certificates)
 | +- CA Root certificate
 | +- CA Intermediate Certificate
 |
 +-+ Certificate Revocation List bundle
   +- Certificate Revocation List A
   +- Certificate Revocation List B
```

The name of the profile would have to be specified through a new string field `ssl_profile` which is added in the `RotateCertificateRequest` type. 
If the `ssl_profile` field is left blank then the gNSI service should rotate the SSL profile which is currently in-use by the gNSI service.  

An example of what this would look like in the certz.proto file:
```protobuf

// Request messages to rotate existing certificates on the target.
message RotateCertificateRequest {
  // If set to `true` the requested operation will succeed even if the
  // `version` is already in use (is not unique).
  // If set to `false` the requested operation will fail and the streaming RPC
  // will be closed with the `AlreadyExists` error if the `version` is already
  // in use (is not unique).
  // It is a no-op for the `finalize_rotation` request.
  bool force_overwrite = 1;

  // An identifier for the specific SSL profile 
  // (collection of certs/bundles/CRLs) 
  // which is being rotated through this stream
  // Leaving this field blank means that this stream
  // will rotate the SSL profile which is currently 
  // being used by the gNSI service.
  string ssl_profile = 2;

  // Request Messages.
  oneof rotate_request {
    GenerateCSRRequest generate_csr = 3;
    UploadRequest certificates = 4;
    FinalizeRequest finalize_rotation = 5;
  }
}
```

[...]

PR #51, opened 2022-11-09
Adds support for SSL profiles to certz. Issue #45

tomek-US:


certz/certz.proto
```diff
@@ -26,33 +26,68 @@ option (gnoi.types.gnoi_version) = "0.2.0";
...
+// |   +-+ Certificate Revocalion List B
 // |
-// +-+ trust bundle (Certificate Authority certificates)
-// | +- CA Root certificate
-// | +- CA Intermediate Certificate
-// |
-// +-+ Certificate Revocation List bundle
-//   +-+ Certificate Revocalion List A
-//   +-+ Certificate Revocalion List B
+// +-+ Another SSL profile used by another gRPC-based service
```
robshakir, line 58: Should we say here that this doesn't need to be a gRPC-based service? e.g., I think https server credentials could be provided here, right?

certz/certz.proto
```diff
@@ -26,33 +26,68 @@ option (gnoi.types.gnoi_version) = "0.2.0";
...
+// the `AddSSLProfile()` RPC. In any case, the assignment of a SSL profile to
+// a service is done using modifying the `/oc-sys:system/
+//  oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/oc-sys-grpc:config/
+//  oc-sys-grpc:certificate-id` OpenConfig leaf.
 //
-// The above shown topology implies that every operation performed using this
-// service changes a credential for all gRPC services at the same time.
+// Note that the this architesture implies that every operation performed
+// on a credential that is part of a SSL profile specified in the request
+// changes the credential at the same time for all  services configured to use
```
robshakir, line 89:
```suggestion
// changes the credential at the same time for all services configured to use
```

certz/certz.proto
```diff
@@ -26,33 +26,68 @@ option (gnoi.types.gnoi_version) = "0.2.0";
...
+// Other services that require credentials _can_ use the same SSL profile as
+// the gNxI server or they can use a SSL profiles that is created using
+// the `AddSSLProfile()` RPC. In any case, the assignment of a SSL profile to
+// a service is done using modifying the `/oc-sys:system/
+//  oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/oc-sys-grpc:config/
+//  oc-sys-grpc:certificate-id` OpenConfig leaf.
 //
-// The above shown topology implies that every operation performed using this
-// service changes a credential for all gRPC services at the same time.
+// Note that the this architesture implies that every operation performed
```
robshakir, line 87:
```suggestion
// Note that the this architecture implies that every operation performed
```

robshakir, line 87:
```suggestion
// Note that the this architecture implies that every operation performed
```

certz/certz.proto
```diff
@@ -185,6 +220,30 @@ service Certz {
   rpc Rotate(stream RotateCertificateRequest)
       returns (stream RotateCertificateResponse);
 
+  // AddSSLProfile is part of SSL profile management and allows for adding new
```
robshakir, line 223: Maybe just `AddProfile`? Not sure whether there are other things that we'd `Add` within this service? (Just wondering whether we can make the naming shorter :-))

certz/certz.proto
```diff
@@ -185,6 +220,30 @@ service Certz {
...
+  // SSL profile.
+  // When a SSL profile is added all its elements, i.e. certificat, CA trust
```
robshakir, line 225:
```suggestion
// When a SSL profile is added all its elements, i.e. certificate, CA trust
```

certz/certz.proto
```diff
@@ -185,6 +220,30 @@ service Certz {
...
+  // bundle and a set of certificate revocation lists are created and they are
+  // empty, so before they can be used they have to be 'rotated' using
+  // the `Rotate()` RPC.
+  // Note that an attempt to add an already exisitng profile will be rejected
+  // with an error.
+  rpc AddSSLProfile(AddSSLProfileRequest) returns (AddSSLProfileResponse);
+
+  // DeleteSSLProfile is part of SSL profile management and allows for removing
+  // an existing SSL profile.
+  // Note that an attempt to remove a not exisiting exisitng profile will result
```
robshakir, line 235:
```suggestion
  // Note that an attempt to remove a profile that does not exist will result

**Review by robshakir — CHANGES_REQUESTED** [review]
This generally LGTM -- will re-ponder on in, but a couple of questions as a first cut.

File: certz/certz.proto [hunk 7]
```diff
@@ -185,6 +220,30 @@ service Certz {
...
+  // SSL profile.
+  // When a SSL profile is added all its elements, i.e. certificat, CA trust
+  // bundle and a set of certificate revocation lists are created and they are
+  // empty, so before they can be used they have to be 'rotated' using
+  // the `Rotate()` RPC.
+  // Note that an attempt to add an already exisitng profile will be rejected
+  // with an error.
+  rpc AddSSLProfile(AddSSLProfileRequest) returns (AddSSLProfileResponse);
+
+  // DeleteSSLProfile is part of SSL profile management and allows for removing
```
marcushines, line 233: does should delete be failed of any service currently has that profile active?
Seems like the other behavior is annoying to deal with in likely you could delete the profile which then leaves in you a race condition of - should this force a reload of the servers which have the profile or do they go into lame duck / handle remaining rpc's then reset?

since that profile no longer exists .. then does the sever go into a failed state? or?

but if we just fail the delete request with "in use" it is very simple

marcushines (commented):
This looks really good thanks Tomek

certz/certz.proto
tomek-US: Yes. I have removed `gRPC-based`

certz/certz.proto
tomek-US: Done

certz/certz.proto
tomek-US: Done

tomek-US: Done

certz/certz.proto
tomek-US: OK. Done

certz/certz.proto
tomek-US: Done

certz/certz.proto
tomek-US: Yeah. That is a very good point but it is a bit in conflict with the fact that those profiles are assigned via gNMI. I can add a paragraph describing that the setting of SSL profile IDs have to be coordinated with gNSI but is this enough?

certz/certz.proto
```diff
@@ -26,40 +26,75 @@ option (gnoi.types.gnoi_version) = "0.2.0";
...
+// Revocation List (CRL) bundles grouped into SSL profiles.
+// Each profile has a unique ID and consists of exactly one certificate, one
+// CA trust bundle and one CRL bundle.
+// There is at least one SSL profile present on the target, the one that is used
+// by all gNxI microservices. It is created during the bootstrap phase and
+// cannot be removed.
+// Other services that require credentials _can_ use the same SSL profile as
+// the gNxI server or they can use a SSL profiles that is created using
+// the `AddProfile()` RPC. In any case, the assignment of a SSL profile to
+// a service is done using modifying the `/oc-sys:system/
```
brianneville, line 83:
```suggestion
// a service is done by modifying the `/oc-sys:system/
```

certz/certz.proto
```diff
@@ -26,40 +26,75 @@ option (gnoi.types.gnoi_version) = "0.2.0";
...
+//  oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/oc-sys-grpc:config/
+//  oc-sys-grpc:certificate-id` OpenConfig leaf.
 //
-// The above shown topology implies that every operation performed using this
-// service changes a credential for all gRPC services at the same time.
+// Note that the this architecture implies that every operation performed
```
brianneville, line 87:
```suggestion
// Note that this architecture implies that every operation performed
```

certz/certz.proto
```diff
@@ -223,6 +288,36 @@ message RotateCertificateResponse {
...
 
+// A request to add one or more SSL profiles.
+message AddProfileRequest {
+  // A list of IDs of SSL profiles to be created.
+  repeated string ssl_profile_ids = 1;
+}
+
+// A response to a request to add one or more SSL profiles.
+message AddProfileResponse {
+}
```
brianneville, line 299: What is the expected handling for the case where multiple IDs are provided here, and one of them fails to be added? 
For example, given ssl_profile_ids of  `[]string{ "foo", "bar", "baz" }`, imagine profile `foo` is created successfully but there is some error with creating profile `bar`.
We are limited here to only return either a single `AddProfileResponse` or an error. 

The error code itself wouldn't suffice to indicate which profiles were added successfully, so the client would have to do parsing on the error description to figure out which profiles were added and which caused errors. 

Two potentially cleaner ways to deal with this are:
1. make `AddProfileResponse` contain a map of the results of the request, similar to https://github.com/openconfig/gnoi/blob/79709cdf28e160f0732ff0086902c06c9c2d74f8/packet_link_qualification/packet_link_qualification.proto#L53
Something like:
```protobuf
// AddProfileResponse will return a map of the status of an AddProfileRequest.
// The map key is the SSL profile ID provided in the AddProfileRequest.
// Valid status responses are:
// OK: SSL profile had been added successfully.
// ALREADY_EXISTS: if the SSL profile ID is already present on the target.
// UNKNOWN: the target cannot add this profile for some unexpected reason.
message AddProfileResponse {
  map<string, google.rpc.Status> status = 2;
}
```

2. limit AddProfileRequest so that it only supports adding a single SSL profile, like:
```protobuf
// A request to add a new SSL profile.
message AddProfileRequest {
  // The ID of the SSL profile to be created.
  string ssl_profile_id = 1;
}
```

certz/certz.proto
```diff
@@ -223,6 +288,36 @@ message RotateCertificateResponse {
...
+
+// A request to remove one or more SSL profiles.
+message DeleteProfileRequest {
```
brianneville, line 302: This has the same issue as the [above comment ](https://github.com/openconfig/gnsi/pull/51/files#r1021579217) regarding `AddProfileRequest`
i.e. its not clear how the target should handle an error if one of the ID's fails to be removed successfully. 

Whatever approach is chosen for solving this issue in the case of `AddProfileRequest` should also be used here for consistency

certz/certz.proto
tomek-US: The (silent) assumption is that the RPC is atomic. If one of the sub-operations fails then the whole RPC fails (with error message saying that a particular profile has not been created).

So, in my opinion in addition to the to options you have mentioned we have third one: improve the documentation.

I think option 2 works for me best.

brianneville: > (with error message saying that a particular profile has not been created).

Parsing the error description to find the cause of the RPC failure in this would create hassle for the client - unless the error description's format was standardized, the client would also have to perform different parsing for each vendor's error description.
 
I think it would be more robust to not need to parse the error description to know exactly which profile ID caused the RPC to fail. 

Option 2 also works best for me, just because it feels like the most simple option (one request for one ssl_profile_id = one response or error)

certz/certz.proto
tomek-US: I have removed the `repeated` keyword. Please taka a look.

certz/certz.proto
tomek-US: I have removed the `repeated` keyword. Please taka a look.

brianneville: the changes look good to me - thanks Tomek!

tomek-US (commented):
All requested changes have been done.

marcushines (approved):

robshakir (approved):

PR #52, opened 2022-11-22
certz: add CSRSuite enums

brianneville:
The possible combinations of KeyType, SignatureAlgorithm and CertificateType (and key_size_bits for RSA) have been consolidated into enums.
This simplifies the request for the client, by hardcoding the types of CSR requests that can be sent.

key_size_bits also had ambiguity for non-RSA keys, which is resolved by including information about the key size in the enum name.

The enum names also specify that SHA-2 is being used, so as to be forward compatible if SHA-3 is adopted in future

This commit also updates the certz.pb.go file to  include these changes


certz/certz.proto
```diff
@@ -223,13 +223,44 @@ message RotateCertificateResponse {
...
+
+  CSRSUITE_X509_KEY_TYPE_ECDSA_SECP384R1_SIGNATURE_ALGORITHM_SHA_2_256 = 13;
+  CSRSUITE_X509_KEY_TYPE_ECDSA_SECP384R1_SIGNATURE_ALGORITHM_SHA_2_384 = 14;
+  CSRSUITE_X509_KEY_TYPE_ECDSA_SECP384R1_SIGNATURE_ALGORITHM_SHA_2_512 = 15;
+
+  CSRSUITE_X509_KEY_TYPE_ECDSA_SECP521R1_SIGNATURE_ALGORITHM_SHA_2_256 = 16;
+  CSRSUITE_X509_KEY_TYPE_ECDSA_SECP521R1_SIGNATURE_ALGORITHM_SHA_2_384 = 17;
+  CSRSUITE_X509_KEY_TYPE_ECDSA_SECP521R1_SIGNATURE_ALGORITHM_SHA_2_512 = 18;
+
+  CSRSUITE_X509_KEY_TYPE_EDDSA_ED25519 = 19;
```
morrowc, line 257: I agree that we should have specified these in an enum as done here not left these
for Stiring string STRING matching :(

morrowc (approved):

tomek-US (approved):
LGTM

PR #54, opened 2022-11-25
authz: clarify Rule name field in README

brianneville:
Clarify that the name field in the Rule message should be unique.

As defined in
https://github.com/grpc/proposal/blob/master/A43-grpc-authorization-api.md#proposal


tomek-US (approved):

morrowc (approved):
This was the original intent, thanks for clarifying in the docs! :)

marcushines (approved):

PR #56, opened 2022-11-28
Add provision to specify the crypt hash

nmahabaleshwar:
Add enum's to specify the crypt hash while rotating the password via gnsi.credentialz


credentialz/credentialz.proto
```diff
@@ -510,11 +510,24 @@ message UserPolicy {
 message AuthorizedUsersResponse {
 }
 
+// Hash type's
+enum HashType {
+  HASH_TYPE_UNSPECIFIED = 0;
+  HASH_TYPE_CRYPT_MD5 = 1;
+  HASH_TYPE_CRYPT_SHA_2_512 = 2;
+}
```
tomek-US, line 518: Is this enum going to be used in other messages than the CryptoHash message? If not then please move it into it.

credentialz/credentialz.proto
```diff
@@ -510,11 +510,24 @@ message UserPolicy {
...
+
+message CryptoHash {
+  // Specifies the hash type used and value
+  HashType hashType = 1;
+  string hashValue = 2;
+}
```
tomek-US, line 524: Is this enum to be used in other messages than the Password message? If not then please move it inside it.

credentialz/credentialz.proto
nmahabaleshwar: Addressed this

credentialz/credentialz.proto
nmahabaleshwar: Addressed this

PR #57, opened 2022-12-01
Add encryption type and description to CA public key

nmahabaleshwar:
Added encryption type and an optional description to CA public key


morrowc:
I think the conflicts here are 're-run protoc' fixed.

morrowc (approved):
This looks ok to me, we should have accounted for key/cert different data requirements originally, whoops.

tomek-US (approved):
LGTM

Issue #58, opened 2022-12-16
Proposal to use the public key generated on the box for creating SSH host certificate


nmahabaleshwar:
Some switches have the capability to generate private/public keys eliminating the necessity to generate them externally and transferring them to the switch. This avoids the potential security risk on transmitting private keys over the wire. Below is a proposal which helps operators utilize this capability.

Below will be the sequence of steps,
1. Query the key generation capability of the switch. This is basically providing a key specification and checking if the switch can generate the public/private key.

	`rpc CanGenerateKey(CanGenerateKeyRequest) returns (CanGenerateKeyResponse)`

2. Trigger generation of the key.

	`rpc GenerateKey(GenerateKeyRequest) returns (GenerateKeyResponse)`

3. Poll to check if the key generation is complete. Key generation takes time, hence we need to poll the switch for key generation. If the key is ready, the public key portion of it should be returned else a simple `Not ready`.

	`rpc GetPublicKey(GetPublicKeyRequest) returns (GetPublicKeyResponse)`

4. Sign the public key with the CA's private key and generate the host certificate.
5. Install the generated host certificate onto the switch.

	`rpc RotateHostCredentials(RotateHostCredentialsRequest) returns (RotateHostCredentialsResponse)`

This is an existing RPC, we can just use the existing message ServerKeysRequest. The public/private key portions of the message need not be set.

If the operator does not wish to use this capability or if the switch does not support this capability, we can directly go to Step 5 and install the externally generated keys and certificates using the existing RPC.

tomek-US:
IMHO there is no need for additional RPCs (perhaps with exception of CanGenerateKey() but I would argue that it also can be part of RotateHostCredentials) - the GenerateKeyRequest/GenerateKeyResponse can be made part of the RotateHostCredentials streaming RPC.

In such case instead of sending ServerKeysRequest you would:
(1) send GenerateKeyRequest and
(2) wait for GenerateKeyResponse  that would contain the generated public key and then
(3) after signing it with CA the next message would be ServerKeysRequest with the signed certificate.

This approach has the advantage that it (1) allows for optionality - the client uses the same RPC but can chose the steps, (2) makes the process a transaction that can consist of rotation of all related credentials.

nmahabaleshwar:
To align with Certz, we can have a separate RPC to query key generation capability and just new message in the existing Rotate RPC for triggering key generation and retrieve it. Below is the updated steps,

[...]

PR #59, opened 2022-12-27
Support for generating public/private key on the box

nmahabaleshwar:
Details of the proposal is in https://github.com/openconfig/gnsi/issues/58


credentialz/credentialz.proto
```diff
@@ -582,14 +585,36 @@ message RotateHostCredentialsResponse {
   }
 }
 
+// KeySuite specifies the valid types of keys and their lengths that
```
morrowc, line 588: this comment doesn't appear to match the content.

credentialz/credentialz.proto
```diff
@@ -582,14 +585,36 @@ message RotateHostCredentialsResponse {
...
+// can be used.
+enum KeyGen {
```
morrowc, line 590: KeyGen is very similar to KeyType (line 397), should they be the same?

credentialz/credentialz.proto
```diff
@@ -582,14 +585,36 @@ message RotateHostCredentialsResponse {
...
+  KEY_GEN_SSH_KEY_UNSPECIFIED = 0;
+  KEY_GEN_SSH_KEY_TYPE_RSA_2048 = 1;
+  KEY_GEN_SSH_KEY_TYPE_ECDSA_P_256 = 2;
+  KEY_GEN_SSH_KEY_TYPE_ECDSA_P_521 = 3;
+  KEY_GEN_SSH_KEY_TYPE_EDDSA_ED25519 = 4;
+}
+
+// A request to ask the target if it can generate key pairs
+message CanGenerateKeyRequest {
```
morrowc, line 599: Can you say why this request/response is necessary?

From the operator side I imagined we would say:
  "Here are your public keys: <set of keys>"
  
 I hadn't seen the need to ask:
  "Can you make an X?"

credentialz/credentialz.proto
morrowc: Trying to keep the ssh key types and possible x.509 key types updated in different enums MAY be useful, but also introduces some accounting problems.

credentialz/credentialz.proto
morrowc: Perhaps this is ssh-keys and cert-keys being crossed up by bad naming choices :(

(though I still imagine you would just dictate the key/cert types to the device)

credentialz/credentialz.proto
```diff
@@ -620,14 +645,24 @@ message CaPublicKeyRequest {
...
+// specification and return the public key back which will be used to create
+// a hsot SSH certificate.
+message GenerateKeysRequest {
+  KeyGen key_params = 1;
+}
+
+// Response from the target whether it can generate the key pairs
+// for the given key specification and the public key if it generated it.
+message GenerateKeysResponse {
+  bool can_generate = 1;
```
morrowc, line 658: There's this can_generate, but also if the message set for CanGenerate is kept, the can_generate is already answered there, right? Why would this need to exist in both places?

credentialz/credentialz.proto
nmahabaleshwar: Will fix the comment

credentialz/credentialz.proto
nmahabaleshwar: We wanted to provide a way for the operator to generate the keys on the box instead of externally generating them, but also give an option to inject externally generated keys. So based on the discussion in https://github.com/openconfig/gnsi/issues/58, we have added some RPC, messages to support this. This API is used to check the key generation capability of the switch.

credentialz/credentialz.proto
erahn: I don't think that we should mix and match KeyGen and KeyType:

- KeyGen specifies, for RSA, the length to be generated as well as the key type
- There could be further divergences in the future.

From talking to [nmahabaleshwar](https://github.com/nmahabaleshwar) out of band, it sounds like there is a desire to merge the SSH and TLS (x509) key types to cut down on the number of enums. I would suggest keeping those separate. They are different protocols and I would not expect the support for the algorithms to be the same across both. Furthermore TLS likes to combine the signature algorithm with stuff like the hash type, which I think we are already proposing to include as a combined bundle.

credentialz/credentialz.proto
nmahabaleshwar: The original intention was to indicate the failure if the requested key can't be generated (The requested key in CanGenerateKey and GenerateKeyRequest are different). We can return an error in such scenarios and don't need this additional field `can_generate`, hence removed it.

credentialz/credentialz.proto
```diff
@@ -620,14 +645,22 @@ message CaPublicKeyRequest {
...
 
+// A request to generate a public/private key pair based on a given key
+// specification and return the public key back which will be used to create
+// a hsot SSH certificate.
+message GenerateKeysRequest {
+  KeyGen key_params = 1;
+}
+
+// Response from the target which has the generated public key.
+message GenerateKeysResponse {
```
tomek-US, line 656: You have to add this to the `RotateHostCredentialsResponse` message at line 584. Otherwise you will not be able to send it over the stream.

credentialz/credentialz.proto
```diff
@@ -620,14 +645,22 @@ message CaPublicKeyRequest {
 message CaPublicKeyResponse {
 }
 
+// A request to generate a public/private key pair based on a given key
+// specification and return the public key back which will be used to create
+// a hsot SSH certificate.
+message GenerateKeysRequest {
```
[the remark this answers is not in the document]
nmahabaleshwar: Done

credentialz/credentialz.proto
nmahabaleshwar: Done

tomek-US (approved):
LGTM

credentialz/credentialz.proto
morrowc: ok, I think I was expecting a data structure that had both the key type and the key size.
It looks like we have to know some magic numbers otherwise.

Admittedly things like ECDSA_P_521 has like 1 key size defined (which happens to be 521 bytes)...
RSA though has ~4 different key types, in this particular case you are specifying: "_2048" in the name.
So for RSA either I keep magic somewhere: "Ah, always 2048 here!" or I have to parse enum text for key lengths.

In the end Tomek seems to approve of this, so that's ok for me :) I am/was probably picking a nit that didn't need pickin'! :)

morrowc (approved):

morrowc:
If you all can't do the merge action here find me on chat and ... I can click the button.

PR #60, opened 2023-01-04
bazel: fix for bazel 6.0.0

brianneville:
bazel 6.0.0 introduced a change which is backward incompatible, see: https://github.com/bazelbuild/bazel/issues/8622

The upb package (@upb) which is a dependancy for rules_proto_grpc (@rules_proto_grpc) had lines in its build which were targeting @bazel_tools//platforms:windows, which had been removed. This made it impossible to build on bazel 6.0.0.
Update the rules_proto_grpc to 4.3.0 to resolve this.

Building with the updated version of rules_proto_grpc would then fail because @com_github_grpc_grpc required c++14 in lots of files. Example error:
external/com_github_grpc_grpc/src/core/lib/gpr/useful.h:109:17: error: use of 'auto' in lambda parameter declaration only available with '-std=c++14' or '-std=gnu++14'

Add a .bazelrc file to force c++14 when building.

Also bump the version of the com_google_googleapis package.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -9,7 +9,7 @@ http_archive(
     urls = ["https://github.com/rules-proto-grpc/rules_proto_grpc/archive/4.3.0.tar.gz"],
 )
 
-# googleapis has not had a release since 2016 - take the master version as of 4-jan-22
+# googleapis has not had a release since 2016 - take the master version as of 4-jan-23
 http_archive(
     name = "com_google_googleapis",
     sha256 = "9fc03150d86501d7da35eefa989d5553bdd77a95cfe4373cdafe8eee92f6bfb1",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,19 +1,20 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 ### Bazel rules for many languages to compile PROTO into gRPC libraries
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -1,19 +1,20 @@
...
-    sha256 = "bbe4db93499f5c9414926e46f9e35016999a4e9f6e3522482d3760dc61011070",
-    strip_prefix = "rules_proto_grpc-4.2.0",
-    urls = ["https://github.com/rules-proto-grpc/rules_proto_grpc/archive/4.2.0.tar.gz"],
+    sha256 = "fb7fc7a3c19a92b2f15ed7c4ffb2983e956625c1436f57a3430b897ba9864059",
+    strip_prefix = "rules_proto_grpc-4.3.0",
+    urls = ["https://github.com/rules-proto-grpc/rules_proto_grpc/archive/4.3.0.tar.gz"],
 )
 
-# googleapis has not had a release since 2016 - take the master version as of 02-nov-22
+# googleapis has not had a release since 2016 - take the master version as of 4-jan-22
```
tomek-US, line 12: 4-jan-23 ?

tomek-US (changes requested):

WORKSPACE
brianneville: whoops, somehow always manage to make that mistake after new years start :)

tomek-US (approved):

PR #61, opened 2023-01-04
certz: add more fields to CSRParams

brianneville:
Add fields allowing clients to specify fields in a CSR.

These fields are ported over from gnoi/cert:
https://github.com/openconfig/gnoi/blob/79709cdf28e160f0732ff0086902c06c9c2d7/cert/cert.proto#L242


certz/certz.proto
```diff
@@ -356,6 +356,16 @@ enum CSRSuite {
 // Used to specify what type of CSR the client expects from the target.
 message CSRParams {
   CSRSuite csr_suite = 1;
+
+  // --- common set of parameters applicable for any type of certificate --- //
+  string common_name = 2;           // e.g "device.corp.google.com"
```
tomek-US, line 361: What about the other fields like V3 extensions (keyUsage, extendedKeyUsage, altName, etc)? Should we add them too?

morrowc: Sometimes (many times?) the process that mints the certs will REQUIRE some of these to be set/unset.. and may NEVER create a cert with these set/unset.

Should we actually make sure that the documentation for cert usage (CPS statement?) for OC/GNSI is clear on what flags to set (and what values are acceptable), and then make sure this list here conforms to that desire?

brianneville: Thanks for the reviews. 
In the next patch i added comments to identify which of the fields are required (only common-name) and which are optional.
Since Arista's products can also support the SAN extension, I added a message type (`V3ExtensionSAN`) to represent the fields for this extension, and added a field with this message type in the CSRParams

tomek-US (approved):
LGTM

Issue #44, opened 2022-10-19
Clarify expectation of trust_bundle with regards to mTLS


RonanMacF:
When confiuring an SSL proile on the switch for mTLS there is generally two thing in particular done

1. A particular CA is marked as trusted. This will be used to authenticate the clients certificate and does not necessarily have to be part of the server certificate chain.
2. A certificate chain is formed on the server with a leaf certificate and 0 or more intermediate certs.

The proto comments contains the following

```
// Target (as seen from gNSI.certificate microservice point of view)
// |
// +-+ certificate (used by all gRPC services)
// | +- certificate (with public key)
// | +- private key
// |
// +-+ trust bundle (Certificate Authority certificates)
// | +- CA Root certificate
// | +- CA Intermediate Certificate
```

This would indicate to me that all CA's as part of the trust bundle are only to be marked as trusted and the server will only have one certificate used in it's chain.

I think a reasonable change to this would be to change

```
  oneof entity {
    Certificate certificate = 3;
    CertificateBundle trust_bundle = 4;
    CertificateRevocationListBundle certificate_revocation_list_bundle = 5;
  }
```

to
```
  oneof entity {
    Certificate leaf_certificate = 3;
    CertificateBundle intermediate_certificates = 4;
    CertificateBundle trust_bundle = 5;
    CertificateRevocationListBundle certificate_revocation_list_bundle = 6;
  }
```

The private key would be optional for all certs in the chain except for the leaf certificate. This is already marked as optional so we are good here.

marcushines:
@tomek-US 

I believe our view was the trust_bundle contains the CA root as well as all intermediates

I am not sure why we would seperate it out like this?

RonanMacF:
It sounds like trust_bundle is the chain including the RootCA that created the leaf server certificate.

This means that there is no option to provide a distinct list of CA's that may have signed a client certificate which would be used in mTLS.

tomek-US:
Yes, the assumption behind this API design decision is that the trust bundle includes the whole chain of CA certificates that are needed to verify the incoming connection from clients. It might be but does not have to be the one that is used to sign the servers certificate. The assumption is that when the server (host) certificate is provided using the Rotate() operation it is from a trusted source and there is no need to verify it.

RonanMacF:
Understood, to a trust_bundle is a list of trusted certificates that the server can use to verify a client during mTLS.

[...]

PR #62, opened 2023-01-10
certz: enforce certificate ordering

brianneville:
Enforce certificate ordering by introducing a CertificateChain type.
When rotating a leaf certificate, this CertificateChain
will be used to specify the chain of certificates being
rotated, up to and including the root cert.

reference:
https://github.com/openconfig/gnsi/issues/44


certz/certz.proto
```diff
@@ -453,8 +453,7 @@ message Entity {
 
   oneof entity {
     Certificate certificate = 3;
-    CertificateBundle trust_bundle = 4;
```
tomek-US, line 456: `trust_bundle` cannot be removed - it is used to rotate the chain of certificates that is used to validate certificate(s) used by the client when establishing the connection. It can be different than the one used by the server certificate that is sent to the client.

brianneville: ack, re-added in the next commit

tomek-US: Thx for adding it back!

Looking at the definition of `CertificateChain` I realized that it is a better version of the `CertificateBundle`. Could you replace `CertificateBundle` with `CertificateChain` for the `trust_bundle` field? And remove the definition of `CartificateBundle`?

brianneville: Done in the most recent patch (https://github.com/openconfig/gnsi/pull/62/commits/4283316122401a71b5fac14f45e7201f7e93f8a0)

tomek-US (approved):
LGTM

PR #70, opened 2023-03-15
Update pathz.proto

marcushines:
This adds a UploadResponse to the RotateResponse to allow signally that the uploadrequest has been processed


tomek-US (approved):
LGTM

PR #80, opened 2023-05-09
Develop new api for getting public key and address the gaps from Arista

marcushines:
I will also look at adding a message to sync both the Cert and Keys in a case for rotation based on new host key generation

### DIFF
--- a/credentialz/README.md
+++ b/credentialz/README.md
@@ -18,7 +18,7 @@ The following files are expected to be created during the bootstrap process:
   * always required
 * target's private key
   * always required
-* `${system_role_home}/.ssh/authorized_users` file for every system account
+* `${system_role_home}/.ssh/authorized_users` file for every system account.  This file contains a list of principals to validate against for access to the system account.
   * always required
   * used to authorize the `username` provided by a client to use this system
     account
@@ -332,6 +332,45 @@ resp := stream.Receive()
 * Check if the new keys and certificate 'work'
 
 * Finalize the operation
+
+```go
+stream.Send(
+    RotateHostCredentialsResponse {
+        finalize: FinalizeRequest {}
+    }
+)
+```
+
+### Rotate Certificate based on existing key.
+
+The most common operation we are expecting to require on devices is the rotation of certificates used for SSH access for devices. This operation expects to reuse the existing host key on the device as there is not really a good reason frequently rotate this identity.
+
+* Get the public key configured on the host.
+
+```go
+
+resp, err := c.GetPublicKey(&GetPublicKeyRequest{})
+```
+
+* Generate certificate basd on key.
+
+* Rotate certificate on device.
+
+```go
+stream.Send(
+    RotateHostCredentialsRequest {
+        server_keys: ServerKeysRequest {
+            certificate: "A....=",
+            version: "v1.0",
+            created_on: 3214451134,
+        }
+    }
+)
+```
+
+* Validate that new settings are working as expected.
+
+* Finalize request
 
 ```go
 stream.Send(
--- a/credentialz/README.md
+++ b/credentialz/README.md
@@ -464,26 +464,6 @@ if _, err := stream.Recv(); err != nil {
 }
 ```
 
-* (Proposal) As there would be a state after the generate key step but before the certficate is returned, the device will not be able to have alignment between the previous cert and new key. We are proposing to add a new state which allows for the atomic rotation of both the key and certificate by holding the rotation of the key until a `Rotate` message is sent after the Response message.
-
-```go
-stream.Send(
-    RotateHostCredentialsRequest {
-        rotate_keys: RotateHostKey{}
-    }
-)
-```
-
-* (Proposal) Validate the Rotate Keys is successful.
-
-At this point both the host key and certificate would be updated to the new values atomically.
-
-```go
-if _, err := stream.Recv(); err != nil {
-    ...
-}
-```
-
 * Validate that new settings are working as expected.
 
 * Finalize request
--- a/credentialz/README.md
+++ b/credentialz/README.md
@@ -208,7 +208,7 @@ stream.Send(
 )
 ```
 
-#### Update the host's keys
+#### Update the host's keys with external keys
 
 * Start streaming RPC call to the target device.
 
@@ -216,14 +216,16 @@ stream.Send(
 stream := RotateHostCredentials()
 ```
 
-* Send a server's keys change request message to the target device.
+* Send a server's keys change request message to the target device. The bytes are expected to be base64 encoded.
 
 ```go
 stream.Send(
     RotateHostCredentialsRequest {
         server_keys: ServerKeysRequest {
-            public_key: "A....=",
-            private_key: "A....=",
+            auth_artifacts: []AuthenticationArtifacts{
+                private_key: []bytes("...."),
+                certificate: []bytes("...."),
+            },
             version: "v1.0",
             created_on: 3214451134,
         }
@@ -231,6 +233,39 @@ stream.Send(
 )
 
 resp := stream.Receive()
+```
+
+* Check if the new keys 'work'
+
+* Finalize the operation
+
+```go
+stream.Send(
+    RotateHostCredentialsResponse {
+        finalize: FinalizeRequest {}
+    }
+)
+```
+
+#### Update the host's keys with generated keys
+
+* Start streaming RPC call to the target device.
+
+```go
+stream := RotateHostCredentials()
+```
+
+* Send a server's keys change request message to the target device. The bytes are expected to be base64 encoded.
+
+```go
+stream.Send(
+    RotateHostCredentialsRequest {
+        generate_keys: GenerateKeysRequest{
+            key_params: KEY_GEN_SSH_KEY_TYPE_RSA_4096,
+        }
+    }
+)
+resp, err := stream.Receive()
 ```
 
 * Check if the new keys 'work'
@@ -382,7 +417,7 @@ stream.Send(
 )
 ```
 
-### Generate new host key and rotate Certificate based on the new key
+### Generate new host key on device and rotate certificate based on the new key
 
 This use case focuses on the rotation of a host key and then generation of the certificate based on the new public key.
 
@@ -391,9 +426,9 @@ This use case focuses on the rotation of a host key and then generation of the c
 ```go
 stream.Send(
     RotateHostCredentialsRequest {
-        generate_keys_request: GenerateKeysRequest {
+        generate_keys: []GenerateKeysRequest {{
             key_params: KeyGen.KEY_GEN_SSH_KEY_TYPE_EDDSA_ED25519 
-        }
+        }}
     }
 )
 ```
@@ -402,7 +437,7 @@ stream.Send(
 
 ```go
 resp, err := stream.Recv()
-data := resp.PublicKey
+data := resp.PublicKeys
 ```
 
 * The caller will then use this data to generate a certificate.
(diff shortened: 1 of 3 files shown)


credentialz/README.md
```diff
@@ -341,6 +341,45 @@ stream.Send(
...
+The most common operation we are expecting to require on devices is the rotation of certificates used for SSH access for devices. This operation expects to reuse the existing host key on the device as there is not really a good reason frequently rotate this identity.
+
+* Get the public key configured on the host.
+
+```go
+
+resp, err := c.GetPublicKey(&GetPublicKeyRequest{})
+```
+
+* Generate certificate basd on key.
```
morrowc, line 355: based

morrowc: 'based on (the|a) key'

credentialz/README.md
```diff
@@ -341,6 +341,45 @@ stream.Send(
 )
 ```
 
+### Rotate Certificate based on existing key.
+
+The most common operation we are expecting to require on devices is the rotation of certificates used for SSH access for devices. This operation expects to reuse the existing host key on the device as there is not really a good reason frequently rotate this identity.
```
erahn, line 346: s/frequently rotate/to frequently rotate/ would be more grammatically correct.

Ending the sentence at "device", as in "This operation expects to reuse the existing host key on the device." would be better advice for general users, imho.

credentialz/credentialz.proto
```diff
@@ -195,6 +195,9 @@ service Credentialz {
 
   // An RPC to ask a target if it can generate a Public/Private key pair
   rpc CanGenerateKey(CanGenerateKeyRequest) returns (CanGenerateKeyResponse);
+
+  // Get the current public key from the host.
```
erahn, line 199: key(s)

credentialz/credentialz.proto
```diff
@@ -697,3 +700,10 @@ message ServerKeysResponse {
 
 message FinalizeRequest {
 }
+
+message GetPublicKeyRequest {
+}
+
+message GetPublicKeyResponse {
+  PublicKey public_key = 1;
```
erahn, line 708: This should be a `repeated` since more than one key may be present. I.e. SSH does one key per algorithm type so this may return 2 values if you have RSA and ed25519 present.

erahn (changes requested):
Marcus, some thoughts below.

credentialz/credentialz.proto
tomek-US: I would vote for adding `KeyType` to both the `GetPublicKeyRequest` and `GetPublicKeyResponse` instead.

credentialz/credentialz.proto
```diff
@@ -195,6 +195,10 @@ service Credentialz {
 
   // An RPC to ask a target if it can generate a Public/Private key pair
   rpc CanGenerateKey(CanGenerateKeyRequest) returns (CanGenerateKeyResponse);
+
+  // Get the current public keys from the host. Each configured key will be
+  // returned in the provided list.
+  rpc GetPublicKeys(GetPublicKeysRequest) returns (GetPublicKeysResponse);
```
tomek-US, line 201: I wonder if this could be made yet another `oneof` message in the `Rotate` stream.

credentialz/credentialz.proto
```diff
@@ -592,7 +597,7 @@ enum KeyGen {
   KEY_GEN_SSH_KEY_UNSPECIFIED = 0;
   KEY_GEN_SSH_KEY_TYPE_RSA_2048 = 1;
   KEY_GEN_SSH_KEY_TYPE_ECDSA_P_256 = 2;
-  KEY_GEN_SSH_KEY_TYPE_ECDSA_P_521 = 3;
```
nmahabaleshwar, line 595: The name of the curve is `P-521` so `KEY_GEN_SSH_KEY_TYPE_ECDSA_P_521` is valid.

credentialz/README.md
```diff
@@ -341,6 +341,126 @@ stream.Send(
...
+### Generate new host key and rotate Certificate based on the new key
+
+This use case focuses on the rotation of a host key and then generation of the certificate based on the new public key.
+
+* Send request for generation of new private key.
+
+```go
+stream.Send(
+    RotateHostCredentialsRequest {
+        generate_keys_request: GenerateKeysRequest {
```
nmahabaleshwar, line 394: `generate_keys_request` has been updated to `generate_keys` in the current review, please update this accordingly.

credentialz/README.md
```diff
@@ -341,6 +341,126 @@ stream.Send(
...
+
+```go
+stream.Send(
+    RotateHostCredentialsResponse {
+        finalize: FinalizeRequest {}
+    }
+)
+```
+
+### Generate new host key and rotate Certificate based on the new key
```
nmahabaleshwar, line 385: Can we mention that the new key is generated on the switch ?

credentialz/README.md
```diff
@@ -341,6 +341,126 @@ stream.Send(
 )
 ```
 
+## User Journeys
```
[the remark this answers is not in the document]
marcushines: we still have the option for sending the private key in - we generate both the private and public remotely and just push them

credentialz/README.md
marcushines: added

credentialz/credentialz.proto
marcushines: up to you?

credentialz/credentialz.proto
nmahabaleshwar: The latest proto file still has `KEY_GEN_SSH_KEY_TYPE_ECDSA_P_512` instead of `KEY_GEN_SSH_KEY_TYPE_ECDSA_P_521`

credentialz/README.md
```diff
@@ -222,8 +222,10 @@ stream := RotateHostCredentials()
 stream.Send(
     RotateHostCredentialsRequest {
         server_keys: ServerKeysRequest {
-            public_key: "A....=",
-            private_key: "A....=",
+            auth_artifacts: []AuthenticationArtifacts{
+                private_key: []bytes("...."),
```
erahn, line 226: Do you want to specify in here the format of this? I.e. base64 encoded openssh keys and openssh ssh certificates?

credentialz/README.md
```diff
@@ -341,6 +343,126 @@ stream.Send(
...
+
+* Validate the `RotateCredentialsResponse`.
+
+```go
+if _, err := stream.Recv(); err != nil {
+    ...
+}
+```
+
+* (Proposal) As there is would be a state after the generate key step but before the certficate is returned the device will not be able to have alignment between the previous cert and new key. We are proposing to add a new state which allows for the atomic rotation of both the key and certificate by holding the rotation of the key until a `Rotate` message is sent after the Response message.
```
erahn, line 434: Nit: `As there is would be a state after the generate key step` I think you should drop the "is" from this statement.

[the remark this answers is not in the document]
erahn: I agree with Niranjan. This breaks from our prior discussion that the credentialz handler would act immediately upon each RotateHostCredentialsRequest message it receives. My understanding at the end of our discussion is that the desired behavior is that GenerateKeysRequest will end with a hostkey being served up, without certificate. The certificate would be added back in by ServerKeysRequest.

credentialz/credentialz.proto
```diff
@@ -592,7 +622,7 @@ enum KeyGen {
...
+  KEY_GEN_SSH_KEY_TYPE_ECDSA_P_512 = 3;
```
erahn, line 625: Explicitly noting that this is still incorrect since the prior conversation is marked as resolved (should be 521, not 512) :)

erahn (changes requested):
Some thoughts and suggestions inline.

credentialz/credentialz.proto
marcushines: had overwritten push

credentialz/credentialz.proto
```diff
@@ -398,7 +427,8 @@ enum KeyType {
   KEY_TYPE_ECDSA_P_256 = 1;
   KEY_TYPE_ECDSA_P_521 = 2;
   KEY_TYPE_ED25519 = 3;
-  KEY_TYPE_RSA = 4;
+  KEY_TYPE_RSA_2048 = 4;
+  KEY_TYPE_RSA_4096 = 5;
```
nmahabaleshwar, line 431: Can you also add `KEY_GEN_SSH_KEY_TYPE_RSA_4096` to the enum `KeyGen` ?

nmahabaleshwar:
You will probably need to run the `regenerate-files.sh` in your local workspace which will generate a new `credentialz.pb.go` and `credentialz_grpc.pb.go` files.

erahn (changes requested):
Marcus, some concerns about your proposal below.

version/version.proto
```diff
@@ -20,5 +20,5 @@ package gnsi;
 
 import "github.com/openconfig/gnoi/types/types.proto";
 
-option (gnoi.types.gnoi_version) = "1.1.0";
-option go_package = "github.com/openconfig/gnsi/version";
\ No newline at end of file
+option (gnoi.types.gnoi_version) = "1.2.0";
```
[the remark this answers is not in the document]
marcushines: is there an implementation of this that matter?

credentialz/README.md
```diff
@@ -341,6 +376,126 @@ stream.Send(
...
+if _, err := stream.Recv(); err != nil {
+    ...
+}
+```
+
+* (Proposal) As there would be a state after the generate key step but before the certficate is returned, the device will not be able to have alignment between the previous cert and new key. We are proposing to add a new state which allows for the atomic rotation of both the key and certificate by holding the rotation of the key until a `Rotate` message is sent after the Response message.
+
+```go
+stream.Send(
+    RotateHostCredentialsRequest {
```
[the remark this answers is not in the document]
marcushines: no it is not optional, the whole point of adding this step is to address your desire to have this as a way to lock step the state of the key and cert.

version/version.proto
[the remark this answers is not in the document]
erahn: My understanding is that there are multiple versions of gNSI running internally on your end. I can't easily answer that question without knowing more about those clients. From our perspective it's not a big deal, we just change our client testing code. But if I was a client and I expected to just worry about if I'm compatible with a major version I'd be unhappy to find out the minor version added a new, mandatory, message.

credentialz/README.md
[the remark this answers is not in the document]
erahn: When we brought this up before we raised it as a concern and you seemed fine with that. To be clear - Arista is fine with having a process where:
- At the start a hostkey / ssh cert are configured
- An RPC connection is opened
- A new hostkey is configured and immediately put into place. The old certificate goes away.
- A new certificate is configured and immediately put into place.
- A finalize command is issued
- The RPC is closed.

When we left the meeting, we thought that was your desired behavior as well. It doesn't quite lead to nice atomic behavior since you can't generate a key on-box and have a new certificate as one operation, but it's not a big deal as long as everyone involved is okay with it. Our goal was to make sure that keys generated on the switch are possible to use for all the flows since the original flow was "keys sent over the wire"-centric.

The problem with having a mandatory RotateIdentity message is that now:
- Do we need this for every flow? If we do then this does feel like a major API change because all the clients need to follow this new flow.
- If we don't then the gNSI server is trying to figure out what flow it is and decide where we are in the sequence. The RPC command is now more stateful than before.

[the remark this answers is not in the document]
marcushines: good then i will delete it literally the only reason i put it in as a proposal was based on your feedback in the meeting

version/version.proto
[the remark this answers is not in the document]
marcushines: there is not workflow change in the this PR - with the removal of the proposal the only new message inserted is generate keys which is just a new option and doesn't break any old functionality.

tomek-US (approved):
LGTM

PR #81, opened 2023-05-10
Adds authentication policy rotation to gNSI.certz API.

tomek-US:


certz/README.md
```diff
@@ -54,6 +54,41 @@ When no longer a profile is needed it can be removed from the target via
 `Certz.DeleteProfile()` RPC. Note that the gNxI SSL profile cannot be
 removed.
 
+#### Authentication Policy
+
+An authentication policy is a set of rules that defines which CAs can sign which
+certificate. By rotating authentication policies, data center admins can ensure
```
danielywong, line 60: what about "which CA can be trusted to sign certificates for which subjects"?

certz/README.md
```diff
@@ -54,6 +54,41 @@ When no longer a profile is needed it can be removed from the target via
...
+that only authorized CAs are able to sign certificates for specific users/roles
```
danielywong, line 61: What about "can ensure all endpoints are updated to validate certificates presented by their peers during mutual authentication are signed by one of the authorized CAs in the authentication framework as specified in the policy"?

certz/README.md
```diff
@@ -54,6 +54,41 @@ When no longer a profile is needed it can be removed from the target via
...
+to all CAs. This root of trust can be used to verify certificates issued by all
+intermediate CAs. In such case, the gRPC server only needs to be configured with
+the “root of trust” certificate. The client must provide a chain of certificates
+that includes certificates from all intermediate CAs.
+
+This organization of certificate issuing processes also allows for
+specialization of CAs. It is possible to limit which CAs can issue certificates
+for particular users. This allows the gRPC server to reject certificates that
+are otherwise valid but were signed by the wrong CA.
+
```
danielywong, line 87: It feels to me that the two paragraphs above are a bit too prescriptive in the context of these APIs.  Suggest to remove.

certz/gnsi-certz.html
```diff
@@ -8433,7 +8473,7 @@ <h1> Module: <font color=blue>gnsi-certz</font>, Namespace:
         <td>current</td>
         <td nowrap>/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/oc-sys-grpc:state/gnsi-certz:counters</td>
         </tr>
-        <tr id="1-1-16-42-65-92-138" class="a">
+        <tr id="1-1-16-42-65-94-138" class="a">
```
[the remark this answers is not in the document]
tomek-US: Those are auto-generated and used in the HTML file only - no need to look at them.

certz/README.md
```diff
@@ -54,6 +54,41 @@ When no longer a profile is needed it can be removed from the target via
 `Certz.DeleteProfile()` RPC. Note that the gNxI SSL profile cannot be
 removed.
 
+#### Authentication Policy
```
ssachinbharadwaj, line 57: Is this authentication policy a TLS standard? If not, where is the spec for this policy semantics? In the proto definition, the policy is of type 'google.protobuf.Any' which does not convey how the policy is represented

certz/README.md
```diff
@@ -54,6 +54,41 @@ When no longer a profile is needed it can be removed from the target via
...
+The solution to this problem is to have a single "root of trust" that is common
+to all CAs. This root of trust can be used to verify certificates issued by all
+intermediate CAs. In such case, the gRPC server only needs to be configured with
+the “root of trust” certificate. The client must provide a chain of certificates
+that includes certificates from all intermediate CAs.
+
+This organization of certificate issuing processes also allows for
+specialization of CAs. It is possible to limit which CAs can issue certificates
+for particular users. This allows the gRPC server to reject certificates that
+are otherwise valid but were signed by the wrong CA.
```
ssachinbharadwaj, line 86: Is this some sort of whitelisting/blacklisting of certain CAs in the CA bundle? If so, why have the wrong CA in the bundle in the first place and then control with an additional authentication policy? Please correct me if my understanding is wrong w.r.t the authentication policy

certz/certz.proto
```diff
@@ -452,6 +479,15 @@ message CertificateRevocationListBundle {
   repeated CertificateRevocationList certificate_revocation_lists = 1;
 }
 
+// An authentication policy.
+// This policy defines which Certificate Authority can sign which certificate.
```
ssachinbharadwaj, line 483: This line is a bit confusing -> "which Certificate Authority can sign". After pushing the CA bundle, any grpc client trying to connect with grpc-server will be validated against the pushed CA bundle. In what workflow does the signing come into the picture after pushing the CA bundle?

certz/certz.proto
```diff
@@ -452,6 +479,15 @@ message CertificateRevocationListBundle {
...
+message AuthenticationPolicy {
+  // Actual authentication policy.
+  oneof policy {
+    google.protobuf.Any serialized = 1;
```
ssachinbharadwaj, line 487: Is this a well-known authentication policy which is accepted via some API by grpc stack or TLS stack (openssl for example)? In other words, is there a way to configure grpc stack to allow or reject the clients based on this policy? Because the grpc stack handles TLS handshake with the clients.

certz/certz.proto
```diff
@@ -480,6 +516,7 @@ message Entity {
     CertificateChain certificate_chain = 3;
     CertificateChain trust_bundle = 4;
     CertificateRevocationListBundle certificate_revocation_list_bundle = 5;
+    AuthenticationPolicy authentication_policy = 6;
```
ssachinbharadwaj, line 519: message UploadRequest contains "repeated Entity". In each Entity, we can have oneof the above 4 items. Is it expected that all the items in the 'entities' list are of different entity types? In other words, can the UploadRequest contain at max 4 'message Entity'?
Though it is possible to create UploadRequest like [certificate_chain, certificate_chain, authentication_policy, authentication_policy] it is wrong. Maybe it can be cleared in the spec that each message in the list must be of a different type. If not, return an appropriate error.

certz/certz.proto
tomek-US: That is the whole idea - tell the gRPC server to accept only a subset of certificates that are signed by a particular CA.

certz/certz.proto
tomek-US: As I have written above the protobuf definition for the description of the policy is being worked on as well as code that will be part of the gRPC framework/generated code that will handle it.

certz/certz.proto
tomek-US: The scenario you have described is possible but while it is strange and seems invalid it there is no problem with it as the last instance of particular type of entity will be made 'active'. 

It is similar to the following piece of C++ code:

```
int a = 0;
a = 1;
a = 2;
```

What counts is that eventually `a == 2` and depending on the implementation the intermediate steps might be ignored and not performed.

Similar effect can be achieved by sending multiple upload request messages over one `Rotate()` stream - what counts is the last one as it will be made 'active' once the `Finalize` message is sent.

certz/README.md
ssachinbharadwaj: What is the field in the client certificate which represents "subjects/users"? Is it the CN (Common Name) or something else? Maybe this will be clear once the policy definition is public as you said.

If we take a step back, I have a question as to why did CA2 sign Alice? Does it mean CA2 "might be" compromised? Since CA2 is part of the trust bundle, should I not trust all the client certificates it has signed? I am trying to understand the use case or security flaw that we are trying to address with this Authentication policy.

certz/certz.proto
ssachinbharadwaj: Thank you. This is clear

certz/certz.proto
ssachinbharadwaj: Ok, this policy needs to be fed to gRPC stack so that it understands the policy and acts accordingly. Makes sense. Once the policy is published publicly and integrated with gRPC framework, can you share the reference of it here? Thanks!

certz/certz.proto
ssachinbharadwaj: Got it. It's more like "client cert 'X' is valid if and only if signed by 'X-CA'". If "client-cert 'X' is signed by 'Y-CA'", it is invalid.

tomek-US: Correct.

certz/README.md
ssachinbharadwaj: Understood. Thanks

certz/README.md
tomek-US: Yes, CA2 might be compromised or there might be a bug in the signing process. In either case such mismatch indicates that there is a problem that needs investigation which might lead to revoking CA2 (or fixing the bug in the process).

certz/certz.proto
tomek-US: Will do.

certz/README.md
ssachinbharadwaj: Ok

danielywong (commented):
Tomek, PTAL.

danielywong (approved):
Tomek, thanks for the work and I don't have further comment except to close of my last feedback with you.

certz/README.md
tomek-US: LGTM. Changed to your version of this paragraph.

PR #93, opened 2023-07-14
Minor clarifications in comments in acct.proto

haussli:


morrowc (approved):

morrowc:
I think you might need to sync this from remote .. or whateverthat's called.

PR #100, opened 2023-07-19
gnsi.accounting.Records are for *all* types of gNxI requests, not just Sets

haussli:


accounting/acct.proto
```diff
@@ -107,7 +107,7 @@ message SessionInfo {
   string remote_address = 4;
   uint32 remote_port = 5;
 
-  // For multiplexing multiple channels over single 5 tuple
+  // For multiplexing multiple channels over a single connection
   // - multiple CLI connections over single SSH
   // - multiple channels in a gRPC conncetion
```
morrowc, line 112: `s/conncetion/connection/`

thanks :(

morrowc (approved):

PR #104, opened 2023-07-26
The default authz policy should allow all RPCs by default until a policy

haussli:
has been set, not permit a vendor to choose the default policy. Implementations should be consistent.

fixes opensconfig/grpc#99

### DIFF
--- a/authz/authz.proto
+++ b/authz/authz.proto
@@ -130,6 +130,7 @@ service Authz {
 
   // Get returns current instance of the gRPC-level Authorization Policy
   // together with its version and created-on information.
+  // If no policy has been set, Get() returns FAILED_PRECONDITION.
   rpc Get(GetRequest) returns (GetResponse);
 }
 
--- a/authz/authz.proto
+++ b/authz/authz.proto
@@ -28,25 +28,50 @@ option go_package = "github.com/openconfig/gnsi/authz";
 //
 
 // Authorization Policy defines which principals are permitted to access which
-// resource. Resources are individual RPC methods scoped by services.
+// resource(s). Resources are individual RPC methods scoped by their path(s).
+// The policy is expressed in JSON format, following the schema in
+// https://github.com/grpc/proposal/blob/master/A43-grpc-authorization-api.md
+// and the protobuf in the
+// README(https://github.com/openconfig/gnsi/blob/main/authz/README.md).
 
-// Example policy:
+// Example UploadRequest with policy:
 //
 // {
+//     "version": "version-1",
+//     "created_on": "1632779276520673693",
 //     "name": "gNSI.ssh policy",
-//     "allow_rules": [{
+//     "policy": {
+//       "allow_rules": [{
 //         "name": "admin-access",
-//         "principals": [
+//         "source": {
+//           "principals": [
 //             "spiffe://company.com/sa/alice",
 //             "spiffe://company.com/sa/bob"
-//         ],
+//           ]
+//         },
 //         "request": {
-//             "paths": [
-//                 "/gnsi.ssh.Ssh/MutateAccountCredentials",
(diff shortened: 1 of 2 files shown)


haussli:
Lint error does make sense; line 34 of authz.proto is 2 characters long and part of comments.  Lines either side of it are <40 chars.

authz/authz.proto
```diff
@@ -28,25 +28,50 @@ option go_package = "github.com/openconfig/gnsi/authz";
 //
 
 // Authorization Policy defines which principals are permitted to access which
-// resource. Resources are individual RPC methods scoped by services.
+// resource(s). Resources are individual RPC methods scoped by their path(s).
+// The policy is expressed in JSON format, following the schema in
+// https://github.com/grpc/proposal/blob/master/A43-grpc-authorization-api.md
+// and the protobuf in the
+// README(https://github.com/haussli/gnsi/blob/main/authz/README.md).
```
ssachinbharadwaj, line 35: Did you mean https://github.com/openconfig/gnsi/blob/main/authz/README.md ?

authz/authz.proto
```diff
@@ -192,6 +217,8 @@ message ProbeRequest {
 // ProbeResponse returns the ACK/NACK for a single user request
 // as evaluated against the current policy, along with the version of the policy
 // that the gRPC call/user were evaluated against.
+// If no policy has been defined, the default response is ACTION_PERMIT, with
+// a zero-length version string.
```
ssachinbharadwaj, line 221: If no policy has been defined, a similar thing has to be defined even for GetResponse right?

For both Probe() and Get(), would it make sense to return FAILED_PRECONDITION error which signifies that the Rotate() has not been performed? Because for both Probe() and Get(), Rotate() is a precondition.

authz/authz.proto
haussli: Oops, yes.   Thanks.

authz/authz.proto
haussli: Is Rotate() a precondition?  Why wouldn't (or shouldn't) Prove or Get work without beginning with a Rotate()?

I do think that you are correct about FAILED_PRECONDITION if no policy has been set yet.

ssachinbharadwaj: What should Get() return when there is no policy?

```
message GetResponse {
    string version = 1;     -> Should it be an empty string?
    uint64 created_on = 2;  -> Should it be 0?
    string policy = 3;      -> Should it be an empty string?
}
```

haussli: I think you misunderstood; I agree with you that if no policy has been defined, Get() should return the error.

But, if no Rotate() has been issued to start a stream, Get() and Probe() should still work, while Get() would still fail if no policy has been set previously.

ssachinbharadwaj: > I do think

Sorry I misread it as "I don't think". My bad

Oh yeah. Get() and Probe() does not have a direct dependency on Rotate() stream.
Yes, I was referring to when the policy is not yet loaded, what should be the Get's behaviour. And I guess we both agree on Get() returning error.

Shouldn't Probe() also return the same error? Just my thought to keep Probe() and Get() consistent if no policy has been set previously.
Will wait for other's opinion

haussli: IMO, Probe() should return what will actually happen if a RPC is executed; so, if a policy has not been defined, it will always return "allowed", rather than being a special case.

morrowc (approved):

authz/authz.proto
morrowc: going three times, resolved.

morrowc:
Please take a crack at resolving the conflicts / sync to head/etc.

haussli:
git does not merge documents very well...

PR #105, opened 2023-07-26
Add SessionInfo.local_address and clarify value of system_address and

### LINKED ISSUE #98: gnsi.accounting.SessionInfo is lacking in various ways
I presume the local/remote fields are for the connection that triggered the accounting message.

system_address should either be local_address, matching local_port, or both system_address and local_address should exist, the latter being the address that the client used as the destination address (which could be any interface on the device).  These fields might hold the same address, which is fine, and clearly having the loopback ("system address") of the device (or VRF) is useful.

What is SessionInfo.layer4_proto?  Is that OSI layer 4, so TCP/UDP/SCTP?  Add comments.

It is not clear what channel_id contains.  Looking at ssh docs/etc, a channel id appears to be an integer; rfc4254 S3.  Is this correct?  The field should be commented appropriately for what the field contains, imiho.

haussli:
layer4_proto.

partially fixes #98

### DIFF
--- a/accounting/acct.proto
+++ b/accounting/acct.proto
@@ -101,8 +101,13 @@ message SessionInfo {
   // on the System which is sending the messages to the remote collector.
   string system_address = 1;
 
-  // Information for 5 tuple for the connection
+  // Information for 5 tuple for the gRPC client connection.  It is possible
+  // for local_address and system_address to be the same, regardless both
+  // must be populated.
+  // layer4_proto is the OSI Transport layer protocol numnber, eg: 6 (TCP);
+  // see http://www.iana.org/assignments/protocol-numbers
   uint32 layer4_proto = 2;
+  string local_address = 7;
   uint32 local_port = 3;
   string remote_address = 4;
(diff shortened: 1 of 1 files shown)


accounting/acct.proto
```diff
@@ -101,8 +101,13 @@ message SessionInfo {
...
   string system_address = 1;
 
-  // Information for 5 tuple for the connection
+  // Information for 5 tuple for the gRPC client connection.  It is possible
+  // for local_address and system_address to be the same, regardless both
+  // must be populated.
+  // layer4_proto is the OSI Transport layer protocol numnber, eg: 6 (TCP);
+  // see http://www.iana.org/assignments/protocol-numbers
   uint32 layer4_proto = 2;
+  string local_address = 4;
```
ssachinbharadwaj, line 110: Field number 4 is already reserved for remote_address

haussli: Grumble.  I keep doing that, forgetting about the field assignment!

morrowc (approved):

PR #109, opened 2023-08-07
Add version and created-on fields for GenerateKeysRequest

nmahabaleshwar:
Summary of changes,
Add version and created-on fields to the GenerateKeyRequest message since that message generates the keys and also adds them to config as host keys. We will need to know when this config rotation happened which will be indicated by the version and created-on fields which are persisted on the switch.


credentialz/credentialz.proto
```diff
@@ -677,7 +677,8 @@ message CaPublicKeyResponse {
 message AuthorizedPrincipalCheckRequest {
   enum Tool {
     TOOL_UNSPECIFIED = 0;
-    TOOL_HIBA_DEFAULT = 1;
+    TOOL_NOT_CONFIGURED = 1;
```
marcushines, line 680: why are you suggesting adding this enum?
what is different from not_configured vs unspecified?
Also what if not hiba would be used for principal check?

credentialz/credentialz.proto
```diff
@@ -688,9 +689,31 @@ message AuthorizedPrincipalCheckResponse {
 
 // A request to generate a list public/private key pairs based on a given key
```
sourcequench, line 690: s/list public/list of public/

credentialz/credentialz.proto
```diff
@@ -688,9 +689,31 @@ message AuthorizedPrincipalCheckResponse {
...
 // specification and return the list of generated public key back which will
```
sourcequench, line 691: s/key/keys

credentialz/credentialz.proto
```diff
@@ -688,9 +689,31 @@ message AuthorizedPrincipalCheckResponse {
 
 // A request to generate a list public/private key pairs based on a given key
 // specification and return the list of generated public key back which will 
-// be used to create a host SSH certificate.
+// be used to create a host SSH certificate. Any existing host keys will be
```
sourcequench, line 692: Generating/rotating keys need not have anything to do with a certificate.

credentialz/credentialz.proto
erahn: Per the protobuf style guide (https://protobuf.dev/programming-guides/style/#enums) UNSPECIFIED seems like a sentinel value to be used as a default, but it does not indicate an explicit configuration request. NOT_CONFIGURED is used to explicitly request that hiba-chk not be used.

nmahabaleshwar: If HIBA is not used we can use authorized principals file for principal check, it can be enabled by
1. `RotateAccountCredentialsRequest{AuthorizedUsersRequest}` to rotate principals for users.
2. `RotateHostParametersRequest{CaPublicKeyRequest}`, this rotates CA public keys and also enables authorized principals file for all users.

marcushines: unspecified is defined as the default - it is up the api design to decide what to do with a default i.e. it is a "valid value" but does require protocol specification on what to do with it

marcushines: you should not redefine tags - just add the new entry as tag 2

nmahabaleshwar: We have removed the proposed new enum and added documentation that `TOOL_UNSPECIFIED` removes any previously configured authorized principals command from configuration.

morrowc (approved):

haussli:
Please change "should be persisted" to "must ...".

PR #113, opened 2023-08-14
Make accounting.SessionInfo.system_address a repeated field that migh…

haussli:
…t include

one "system address" per address family.

### DIFF
--- a/accounting/acct.proto
+++ b/accounting/acct.proto
@@ -100,7 +100,7 @@ message SessionInfo {
   // loopback 0).
   // One per-address family (AF) might exist, all of which should be included
   // in this field.
-  repeated string system_address = 1;
+  repeated string system_addresses = 1;
 
   // Information for 5 tuple for the gRPC client connection.  It is possible
   // for local_address and system_address to be the same, regardless both
--- a/accounting/acct.proto
+++ b/accounting/acct.proto
@@ -93,10 +93,14 @@ service AccountingPush {
 // SessionInfo is the session identifier that identifies the context
 // in which service was executed.
 message SessionInfo {
-  // system_address is the IP address of the System sending the event,
-  // in most cases this address is the Loopback or Loopback equivalent
-  // on the System which is sending the messages to the Collector.
-  string system_address = 1;
+  // system_address is the IP address(es) of the System sending the event.
+  // In most cases this address is the Loopback, or Loopback equivalent,
+  // on the System which is sending the messages to the Collector, and might
+  // be specifically configured for gNxI (eg: gnxi source-address interface
+  // loopback 0).
+  // One per-address family (AF) might exist, all of which should be included
+  // in this field.
+  repeated string system_address = 1;
 
   // Information for 5 tuple for the gRPC client connection.  It is possible
   // for local_address and system_address to be the same, regardless both


morrowc (approved):

accounting/acct.proto
```diff
@@ -93,10 +93,14 @@ service AccountingPush {
...
-  // on the System which is sending the messages to the Collector.
-  string system_address = 1;
+  // system_address is the IP address(es) of the System sending the event.
+  // In most cases this address is the Loopback, or Loopback equivalent,
+  // on the System which is sending the messages to the Collector, and might
+  // be specifically configured for gNxI (eg: gnxi source-address interface
+  // loopback 0).
+  // One per-address family (AF) might exist, all of which should be included
+  // in this field.
+  repeated string system_addresses = 1;
```
earies, line 103: This is better than previous but there is still some ambiguity that needs to be addressed I feel

* What about when no loopback addresses are provisioned?  What would this be populated with if anything?
* A loopback (or any interface) may carry multiple IPv4/IPv6 addresses so are we now factoring in default address selection techniques which can differ by implementation or grabbing and sending all addresses which can include a variety of internal addresses?
* Does `gnxi source-address interface loopback 0` really apply here?  This is a dial-in service in which the L3 headers are intact by way of the underlying gRPC/HTTP/TLS/TCP connection.  So rather than a "source address", this is more so about an identity address.  I could understand a config knob to _pin_ the system addresses which may very well be more predictable than relying off address selection of an interface (config schema knobs like this could be proposed per gRPC service)
* Since it's really about encapsulating the identity of the system sending the messages, this is already known in packet headers.   Is this additional overhead necessary in each message struct?

haussli: hey, ok.
1) I did not mean to imply that it had to be a loopback.  there are various ways to choose the system address.  IIRC, cisco chooses by configuration, loopback, or highest address of all active interfaces, else possibly a default such as 192.168.1.254.
2) I'm not sure that I am following; typically only one address per AF is chosen.  is there another method?
3) yes, this is the idea; some config knob that pins the system address for a service.
4) it is not necessarily know by the packets, as often any address on the device can be used.  it may or may not have dns records for that address.

I'm totally open to your suggestions, and surely the maintainers are too.

earies: >     1. I did not mean to imply that it had to be a loopback.  there are various ways to choose the system address.  IIRC, cisco chooses by configuration, loopback, or highest address of all active interfaces, else possibly a default such as 192.168.1.254.
>     2. I'm not sure that I am following; typically only one address per AF is chosen.  is there another method?

I'll combine (1/2) because they are related to the same concern that there is reliance on a given system's default address selection mechanism for the system identity (which could differ from implementation to implementation)

Yes - 1 per AF still stands but it wasn't clear by the field comments that there was a reliance off the default address selection mechanism of each implementation.  If that is true then this field is only packed w/ > 0 and <= 2 addresses and can be from any interface on the system should a loopback not be configured (or even something non-routable if a system defaults suggest)

>     3. yes, this is the idea; some config knob that pins the system address for a service.

This could then be an override for the default address selection mechanism from above

>     4. it is not necessarily know by the packets, as often any address on the device can be used.  it may or may not have dns records for that address.

With a dial-in service, the client is calling into a NE that it is aware of and has an associated identity.  If we draw the parallels to all other gRPC services under the OpenConfig umbrella, there are not additional message fields in the payload indicating the identity for each RPC response message.  I think this same statement could be said of gNMI `SubscribeResponse`  and if we consider the scale just like we could have here for accounting records, we are then introducing overhead that adds up very quickly.

For gNMI subscription responses, it is up to the client to "tag" the associated identity before marshaling downstream.  It is also very likely that these gRPC dial-in connections are most always targeted at the system identity which would hold DNS records.

But I understand the concern which is probably more predominant in dial-out scenarios as was in TACACS+/RADIUS/SYSLOG, etc...

> I'm totally open to your suggestions, and surely the maintainers are too.

Thx! FWIW on this one, I'd probably mostly lean towards (4) because the scenario is always dial-in, taking into account reducing any overhead where necessary and providing consistency across gRPC services under the various umbrellas.  The client knows which element it has issued this RPC against.

earies:
> Mon, Aug 14, 2023 at 04:52:59PM -0700, Ebben Aries: ...
> This could then be an override for the default address selection mechanism from above
> Does this address 1 - 3? [haussli@198b819](https://github.com/haussli/gnsi/commit/198b819166549f27976701942b7adecaeff639da)

LGTM if kept

> 4. it is not necessarily know by the packets, as often any address on the device can be used. it may or may not have dns records for that address. With a dial-in service, the client is calling into a NE that it is aware of and has an associated identity. If we draw the parallels to all other gRPC services under the OpenConfig umbrella, there are not additional message fields in the payload indicating the identity for each RPC response message. I think this same statement could be said of gNMI `SubscribeResponse` and if we consider the scale just like we could have here for accounting records, we are then introducing overhead that adds up very quickly.
>> It is calling something, which might be the device or a proxy; could it also be a distributed event system like kafka? 
Inbound calling, there is still some identity or mapping table kept _somewhere_ even if not direct connect

On reception, there are 2 solutions to this problem as I see it

1. The ultimate RPC caller needs to add additional data (or metadata) to the data received.  In fact at this stage is possibly where other data manipulation occurs of possibly flattening message structs down to k/v pairs, etc.. to then marshal downstream to message bus, databases, etc..  The "additional" data could be items like hostnames, origins, etc... (e.g. something to distinguish downstream)
2. Extend message structs in the payload to encode this information in every message 

(1) is more efficient and already necessary as I see it while (2) introduces overhead and duplication of design across _all_ services for consistency

> FWIW on this one, I'd probably mostly lean towards (4) ... ... reducing any overhead
>>  But, your comment about other grpc services appears to be true; no others appear to have a similar field. Are you advocating for removal of the field entirely?  

That's where I'm leaning on this one.  gNMI subscriptions already need to take into account (1) above as each message does not reiterate the identity of the producing node.  Extension metadata could do this today but at the expense of unnecessary overhead on what can be a very chatty channel

>> The overhead seems inconsequential to me, if provides a benefit. So, is there a benefit?

If we look across OpenConfig data-modeling and API services as a whole and introduce use-cases and multi-dimensional scale on top, I think we are all going to be in for an awaking very soon where overhead considerations need to be baked in at every level of the design where we need to balance must haves vs. helpers, etc...  For accounting records across all actions that can be invoked across many services, this too can prove overwhelming (considering P4RT, gRIBI, etc..)

haussli:
morrow and i discussed removal of system_address(es) entirely.  We, I believe,
agree to trim that phat.

Would you draft that change?  or shall i?

PR #115, opened 2023-08-16
Refactor accounting service to "acctz"

earies:
- Refactor acctz service including package/service/rpc/message
  nomenclature for consistency
- Remove dial-out "Push" service/RPC

__Proposal: Refactor 'accounting' to 'acctz' and remove "dial-out" service__

This PR is a proposal for the following:

* Refactor accounting services to 'acctz' for consistency to other gNSI
  services
  * Adjust nomenclature to match the above refactoring
  * Rename RecordStream() RPC -> RecordSubscribe()
  * Rename RecordSync{} message -> RecordRequest{}
    * RecordRequest could very well include future parameters such as "filters" so it could expand beyond a sync/timestamp request
  * Rename Record{} message -> RecordResponse{}
* Remove `AccountingPush` service entirely

## Justifications

The refactoring above is a suggested first step to align consistency across
gNSI services.  Currently no behavior differences are introduced to the dial-in
services (no changes to the message structs or RPC inputs/outputs).

The removal of the `AccountingPush` service is suggested as this would imply
"yet another dial-out mechanism specific to a service" and this can be covered
by way of the `grpctunnel` project if absolutely necessary.

https://github.com/openconfig/grpctunnel

Merely reversing the message semantics also requires a separate collector
implementation coupled with configuration schema that would be necessary
(undefined) for parameters such as the destination collector, authentication
criteria, retry behaviors, TLS, etc..

Suggestion is to consider the future but start small on must haves and build up
features over time.  Defining prematurely will likely miss the mark if there
are no use-cases but will give a false signal of stability if coupled with
functional in-use services.

_NOTE: Generated stubs/versioning can be updated in this PR if agreed upon_


earies:
@morrowc @haussli for comment/review

morrowc (commented):
Generally this proto isn't really used yet, so any change seems ok :)
I'm not partial to the old/original plan, but i do like consistency? :)

heas@ has been doing a ton of reading through this and suggesting updates over the last week or a bit, I'd see what his thoughts are, before jumpnig along to the end.

morrowc:
Ah also, I totally support the trim-the-phat effort here, and pointing out; "Hey, you built a swiss-army-knife, but we just wanted a knife?" (yagni) is totally on target. :) I think heas@ has some comments, too, though.

morrowc (approved):

morrowc:
> I have only one complaint with this proposal, 'z'. I do not understand the 'z' thing; I take it to be an attempt at cuteness by abbreviation and replacing an 's' (credentialS, ..) with z. In which case, I would assume "acctz" meant "accounts". I'd leave the name alone. But, I really do not care and I'm not the maintainer here.


I think the 'z' at the end is a ... google-ism... it's fine here, and it does follow the certz/credentialz/etcz things...
so, "for consistency!'

earies:
> > I have only one complaint with this proposal, 'z'. I do not understand the 'z' thing; I take it to be an attempt at cuteness by abbreviation and replacing an 's' (credentialS, ..) with z. In which case, I would assume "acctz" meant "accounts". I'd leave the name alone. But, I really do not care and I'm not the maintainer here.
> 
> I think the 'z' at the end is a ... google-ism... it's fine here, and it does follow the certz/credentialz/etcz things... so, "for consistency!'

yep!  varz, etc... concepts brought into some OSS like prometheus (https://sre.google/sre-book/practical-alerting/)

PR #121, opened 2023-08-18
Trim undefined behaviors from acctz

earies:
This PR is to start discussion on various ambiguous or undefined behaviors.

Choosing to do this as a PR vs. issue to adjust as we discuss.  Will tag items
as inline comments post submission and if we come to consensus on items, will
adjust the final commit message for a squash merge.

### DIFF
--- a/acctz/acctz.proto
+++ b/acctz/acctz.proto
@@ -99,34 +99,30 @@ message SessionInfo {
 // Details of authentication - for gRPC or shell/vendor-CLI.
 message AuthDetail {
   // Identity string is used to identify the user that executed the
-  // command. For instance, it could be the spiffe-id in the case of
+  // command. For instance, it could be the SPIFFE-ID in the case of
   // gRPC or unix-style user-name in the case of shell/vendor-CLI.
   string identity = 1;
 
-  // Privilege level configured on the system.
-  uint32 privilege_level = 2;
-
-  // authentication status
+  // Authentication status
   enum AuthenStatus {
     AUTHEN_STATUS_UNSPECIFIED = 0;
     AUTHEN_STATUS_PERMIT = 1;
     AUTHEN_STATUS_DENY = 2;
   }
-  AuthenStatus status = 3;
+  AuthenStatus status = 2;
 
   // In case of STATUS_DENY, cause for the deny
-  string deny_cause = 4;
+  string deny_cause = 3;
 }
 
 // Command details for shell/vendor-CLI
 message CommandService {
   enum CmdServiceType {
     CMD_SERVICE_TYPE_UNSPECIFIED = 0;
-    CMD_SERVICE_TYPE_SHELL = 1;
-    CMD_SERVICE_TYPE_CLI = 2;
-    CMD_SERVICE_TYPE_HTTP = 3;		// Web UIs
-    CMD_SERVICE_TYPE_RESTCONF = 4;
-    CMD_SERVICE_TYPE_NETCONF = 5;
+    CMD_SERVICE_TYPE_CLI = 1;
+    CMD_SERVICE_TYPE_HTTP = 2;		// Web UIs
+    CMD_SERVICE_TYPE_RESTCONF = 3;
+    CMD_SERVICE_TYPE_NETCONF = 4;
   }
   CmdServiceType service_type = 1;
 
@@ -210,10 +206,6 @@ message RecordResponse {
 
   // Authentication related details
   AuthDetail authen = 7;
-
-  // Optional repeated task_id that represent tasks that were used to
-  // accomplish the request on the system.
-  repeated string task_ids = 32;
 }
 
 // RecordRequest, requests a starting point for records to be sent to the


acctz/acctz.proto
```diff
@@ -99,34 +99,30 @@ message SessionInfo {
 // Details of authentication - for gRPC or shell/vendor-CLI.
 message AuthDetail {
   // Identity string is used to identify the user that executed the
-  // command. For instance, it could be the spiffe-id in the case of
+  // command. For instance, it could be the SPIFFE-ID in the case of
   // gRPC or unix-style user-name in the case of shell/vendor-CLI.
   string identity = 1;
 
-  // Privilege level configured on the system.
-  uint32 privilege_level = 2;
```
earies, line 107: Privilege levels are custom to only some implementations and generally sit between the range of 0-15 (16 values) with a few system defined levels that carry similar behaviors leaving room for user-defined levels and privilege mapping.

Now, from this I would like request for comment on the following:
* Does this even need to be sent for every accounting record?  User permissions mappings to resources are abstract of this privilege level custom to only some implementations.  Others may use other techniques such as user classes/groups and that is not conveyed here
* If it becomes optional for implementations that do not support then we have an issue omitting the field as it would take on the assumption of the value 0 for protobuf (which can be mixed in with an actual value).  One option would be to always set something like 2^32-1 (4294967295) in such cases but that is also unnecessary overhead

acctz/acctz.proto
```diff
@@ -99,34 +99,30 @@ message SessionInfo {
...
   // In case of STATUS_DENY, cause for the deny
-  string deny_cause = 4;
+  string deny_cause = 3;
 }
 
 // Command details for shell/vendor-CLI
 message CommandService {
   enum CmdServiceType {
     CMD_SERVICE_TYPE_UNSPECIFIED = 0;
-    CMD_SERVICE_TYPE_SHELL = 1;
```
earies, line 125: Underlying "shell" accounting to my knowledge is not implemented across any shipping network implementation today so I guess I would like to bring this one back to "how" this would be suggested, what this might entail first before adding a variant here which cascades to the remainder of other messages/field options here.

Authorization and accounting of supported NBI is very well supported but since most systems implement on top of some alternate base OS, the underlying OS shell capabilities (either control or forwarding plane) are not subject to the same AAA.  Generally this stops at if a user can access the shell or not coupled with an underlying group mapping that may or may not allow privilege escalation (su/sudo).

Each implementation could support a number of underlying shells (csh, sh, bash, zsh, etc..) and while there are techniques to "intercept" and log commands via shell profiles, this could likely be overridden by a user and would also call for those subsystems to be able to hook into accounting subsystems outside of general logging (file/syslog)

I fully comprehend that this means that there is lack of visibility into what a user that has these privileges may do in an underlying shell but maybe we can bring this back to if or how this is done in the compute space first and then reintroduce capabilities back into gNSI once defined.

Accounting is also the only service currently concerned with the underlying shell.  Authorization would be an entirely other conversation as well.

acctz/acctz.proto
```diff
@@ -210,10 +206,6 @@ message RecordResponse {
 
   // Authentication related details
   AuthDetail authen = 7;
-
-  // Optional repeated task_id that represent tasks that were used to
-  // accomplish the request on the system.
-  repeated string task_ids = 32;
```
earies, line 216: My understanding off this 2-line comment would be that this could represent any system defined method to "trace" a call.  Could you elaborate on it's intent or shall we mark for removal and define what this might mean later?

acctz/acctz.proto
```diff
@@ -99,34 +99,30 @@ message SessionInfo {
...
-  AuthenStatus status = 3;
+  AuthenStatus status = 2;
 
   // In case of STATUS_DENY, cause for the deny
-  string deny_cause = 4;
+  string deny_cause = 3;
 }
 
 // Command details for shell/vendor-CLI
 message CommandService {
```
earies, line 119: We have a split of gRPC and "command" services however the command services listed here are a mix of separate APIs and each API behavior may or may not be able to be broken down into the constructs of commands and command arguments.

For CLI, it might be natural to break this down in such a fashion

However for other NBI APIs you have a mix of protocol level semantics coupled with data payloads in various structures, some of this is analogous to gRPC services where currently the plan is to encapsulate the entire message in a PROTO_ANY

I think this will need some additional consideration if we were to elaborate with examples on all services.  Currently in JUNOS, we log/record across various APIs by mostly breaking it down at the RPC name level only but not with full visibility into the remainder of request.

The full requests are analogous to debug/tracing which come with additional cost

acctz/acctz.proto
```diff
@@ -179,6 +175,9 @@ message GrpcService {
 
   // True, if truncation of payloads occurs due to an implementation
   // limitation in the originating service, any middleware, or the receiver.
+  // Truncation must only occur within a given API's message field (non
```
earies, line 178: Adjusted wording here to clarify that any truncation should only occur within the contents of a protobuf message field and not on the message/field boundaries themselves to allow for proper deserialization.

But I would like to raise the point that this is going beyond where we have been in classical accounting and is more analogous to full debug/tracing.  Since this is crossing such boundaries, I would likely propose additional configuration schema (or runtime gRPC APIs - however this is also going to be different among implementations) to toggle if this behavior should be enabled/disabled and not necessarily _always on_

acctz/acctz.proto
haussli: I was thinking about this separately.  priv-lvl is not universal; other platforms use user-defined names to identify roles, such as "operator".  As such, I think this should be a string, not an int.  It would be more versatile that way and could still hold a number.
It could be moved, but I think some parity with existing accounting is useful and do not consider this overhead.

acctz/acctz.proto
haussli: It seems like command accounting to me.  cli command accounting includes the command and args....  This is just more volumous.

acctz/acctz.proto
haussli: You seem to be conflating expanding the list service_requests with this log message?

I do not know what NBI is.

I do not see how the junos behavior you describe is different from the GrpcService message.

acctz/acctz.proto
haussli: I think these are intended to be IDs that could be used to correlate processes etc with an RPC/cmd/etc.   I believe that some systems include such IDs in syslog msgs.  But, I was not around when this was created, so I am speculating.

acctz/acctz.proto
haussli: None of that means that this accounting is not useful or could the issues with the method could not be overcome.  I do not see a reason to remove this.

earies: The reason why I would suggest removal and trimming such behaviors is because its currently an undefined wish list item is all.  A protobuf IDL with fields and comments is not the same as a specification that considers how to implement and will result in misperception and ambiguity and thus differences and partial support among implementations.

At a high level, I do not disagree to the usefulness because without such, there is loss of visibility into some actions that are performed by already trusted classes of users however...

A network element is mostly a distributed system wrapped up to abstract a single management interface - there are fully supported interfaces into the network element and then there can be many non-supported interfaces/debug shells such as:

* Underlying controller card OS shells
* Linecard OS/ukern shells
* Other components, SoC - BCRM debug shell, etc...

In that means there are various methods to essentially "jump" or "attach" to such endpoints by breaking out of the supported interface - it's many "systems" internally.  Some but not all capabilities are brought up to the fully supported interfaces (e.g. common and absolutely necessary often providing some abstraction) - these generally inherit the supported AAA infra.

Each one of these interfaces has vastly different underlying resources and may or may not be modifiable from implementation to implementation so I'd like to peel this one back a bit to what this might look like and what might be realistic when we're talking about accounting to various (sub)systems which also as I mentioned would also really apply to any vision of authorization as well.  This problem statement need not apply to network elements either and would pose a similar issue in various compute host scenarios as well.

So agree that it's useful but would start at dissecting the problem statement w/ various scenarios and then talk solutions before adding it to the API definition as that's the easier of all above.

acctz/acctz.proto
earies: Similar to other cases, if there is no specification or concrete intent considering how such would be implemented consistently, etc..  then I would suggest it's removal to be reintroduced at a later point in time should there be a need and detailed specification.

acctz/acctz.proto
haussli: My point is that it exists today, though it has the caveat that you mention. But the caveat exists not be cause it is not possible to avoid, but due to the laziness of the implementation.

acctz/acctz.proto
haussli: Marcus & Morrow should comment on this; they are more likely to know the intention.

acctz/acctz.proto
haussli: You are suggesting splitting netconf/restconf in its own service_request type, because they more closely resemble (g)rpc and their arguments might not be representable in the string of cmd_args.  That seems reasonable to me.

For http, what I've seen of these interfaces, they are just conduit of sorts to the cli.  I realize that might not be universal and it could carry binary request data.  I suppose it co-exist with netconf, or be its own type.  Binary data could also be base64; which seems common.

I  do not know what any of this has to do with the deny_cause field. :)

earies: The comment is tagged to the `message CommandService` line but review rendering here makes it look like its against the changed line - Since this is a PR, not every line can hold a comment but rather requires a change within proximity (I had considered opening this for discussion w/ 1 or more distinct issues but opted for 1 PR to open up a wider discussion).  `deny_cause` has nothing to do w/ this but rather just had a alignment in field numbers within proximity.

That aside, I do think each distinct API will need need further analysis that is likely to be conveyed as their own message structs otherwise an additional ruleset would need to be devised on top of how to appropriately map, what will not map, etc...

Since none of that is defined today, it's just an additional point that there are versions of gNSI being cut (currently tagged as `1.2.2` at time of this comment) and implementation underway where there is too much ambiguity which is likely to create a significant amount of experimental backwards incompatible churn.  My suggestion is to eliminate such cases (removal) and build back up as time needs to be spent up front on the dissection of problem statements and specification vs. IDL.

acctz/acctz.proto
earies: I think this goes back to the intentions and goals of gNSI - being a redefinition of AAA, it seems there is a mix of items here that are not covered in their (OpenConfig) YANG model or other gNSI service equivalents (authorization/authz/pathz) but maybe rather map back to existing per-vendor implementation behaviors.

A "user" is bound to a privilege-level internal to a given implementation but this is agnostic to the binding by way of gNSI (and what seems like for good reason).  Correct me if I'm wrong but this "binding" should be known, can be queryable/streamed or its own structure table (aka. AuthorizationPolicy) so does every accounting record need to repeat this?

The answer could very well be yes but to draw the analogy to the gNMI/YANG world, this is like separate views of the same data just organized differently and when looking across all OC projects in tandem - there is an explosion of data by way of modeling across various APIs to be implemented on an element with finite resources.

Just to elaborate on my concern of "overhead", its with respect to a view across the board on what is ultimately protocol design so every bit will count - this is by no means a field that will be the offender but my concern would currently be targeted at full payload reflections in either success/failure scenarios especially when dealing with very large config sets (Coupled with OpenConfig modeled expansions that are likely to magnify a data payload an order of magnitude from previous as items like apply-groups, macro expansions, etc.. are not used)

acctz/acctz.proto
earies: Could you clarify the last statement?  Are you saying command accounting (and/or authorization) for all of the underlying shells/subsystems specified above does exist today?  I had asked in the initial comment if or how this might be done in the compute space because the analogy can be drawn to any OS and not just captive shells.

It seems that this can be overcorrected and solved by not giving access to the subsystems by any public API (CLI in this bucket as well) but that would be counterproductive in reality for everyone.

IMO its a problem statement that needs brainstorming and collaboration on solutions and not just a field in an API declaration

The intent of the PR is to trim undefined behaviors (which this is) with possible reintroduction back at a later point

acctz/acctz.proto
earies: Yes - similar and likely varied today across various NBIs - CLI command accounting I would venture to bet is the most consistent whereas other APIs may not glean the full payloads or the complete wrapper framing w/o additional debugging/tracing turned on

gRPC message sizes as well as internal structures may end up also gating what comes in via 1 NBI may not easily be reflected out another and with truncation being the middle ground, would a partial message suffice in all cases?  Guessing probably enough to hope to correlate the transaction with the actual client side that invoked if tracing/accounting is performed there as well

acctz/acctz.proto
morrowc: If the 'user' here can change their privilege level during a session/connection I think you need to track that.
In one example. this is:

  ```login as me, enable to do RW things, logout```
  
if you only noted the role/priv-level/etc at the start/end you'd be missing any transitions.
'not an int' seems like a good idea though.

acctz/acctz.proto
morrowc: "for all of the underlying shells/subsystems specified above does exist today?"

yes, for one platform we use, it does...

Is the problem you (ebben) see:
   "Today no platform will be able to account for 100% of user actions, so having this toggle in proto is not helpful"
   (this is your aspirational comment I think?)
   
I suppose that's a fine argument, but what if we open FR's with all vendors we deal with to say:

  "Hey, all user input on the cli MUST be accounted, full stop."

yup, we'll be waiting a while but.. that really is the long term goal, I think.
we COULD also just go add the enum / etc at the time that the features start to be realized.

acctz/acctz.proto
morrowc: Today we get this sort of notation in tac+ accounting logs, it's what stitches together all commands/etc sent over the life of a user's "session".

That's the intent of this field.
That may not matter for gRPRC things - 'connect send config, disconnect' - session-id is same as singular record.
It does matter for user-that-sshs sessions though. (I think it does, still matter)

acctz/acctz.proto
morrowc: The purpose of 'commandservice' is actually to account for 'chris uses ssh/telnet to connect to a  device and types things'.

In a world where folk use some gNMI for config management (for instance) and gNOI for 'collection of stats/etc' MOST interaction is likely robots not humans. Where a human may still connect: "why is this bgp session not coming up?" "Why did this interface not do what I wanted?" "Oh, great, vendor bug, now I have to type 12 'test crash' type commands :("

we want to have the accounting data for that come over the gNSI path.

acctz/acctz.proto
haussli: And, because it can change and because accounting records are not necessarily received/retrieved while the user is still logged-in or even at the same priv-lvl, it might not be possible to retrieve the level in use at the time of an event.

acctz/acctz.proto
haussli: There is one impl. of grpc accounting (non-gnsi) and one of its services does truncate under certain circumstances, and not in the way your change describes.  I do not know how difficult it would be to implement this change, but it does seem useful to me.

Also, whether args are accounted could be controlled by a configuration knob.  or, control the size of args; or both.

morrowc: > Adjusted wording here to clarify that any truncation should only occur within the contents of a protobuf message field and not on the message/field boundaries themselves to allow for proper deserialization.

btw, I believe this really was the original intent, "do not make a pile of garbage when you decide not to send the 10gb config over this RPC, please trim it sanely, kthxbi!"

acctz/acctz.proto
haussli: This field should just be a string.  Rename it to something generic, allowing it to cover the various implementations: privilege_level, user class, user role group, etc etc.  something like

// user's privilege level, user class, user group, task group, etc.
string group;

acctz/acctz.proto
haussli: These non-command-like services should be split into a different (new) service-type.

acctz/acctz.proto
haussli: I'll add this to PR #122

acctz/acctz.proto
nmahabaleshwar: @morrowc You mentioned this field matters for SSH access, it can be the pid of the sshd instance serving the connection ?

morrowc: Could be ssh-pid, sure!
I think I figured the universal answer: "uuid" would work here too.
but yup, sshd pid seems ok too.

morrowc (approved):

acctz/acctz.proto
```diff
@@ -99,34 +99,30 @@ message SessionInfo {
...
-
-  // authentication status
+  // Authentication status
   enum AuthenStatus {
     AUTHEN_STATUS_UNSPECIFIED = 0;
     AUTHEN_STATUS_PERMIT = 1;
     AUTHEN_STATUS_DENY = 2;
   }
-  AuthenStatus status = 3;
+  AuthenStatus status = 2;
```
rnhaddad, line 112: for backwards compatibility, perhaps it is best to NOT change this value? Use [deprecated=true]; ?

rnhaddad: Comment applies throughout this PR

PR #122, opened 2023-08-21
separate authentication status and user info.  Make both members of

haussli:
SessionInfo.
Add a tty field to SessionInfo for the tty/vty; eg: con0, aux0. Add a session status field to SessionInfo to track login/out (aka start/stop records) and requests that are altogether a login, process request, logout events.
Add a AuthorDetail field to RecordResponse for the authorization of a request.

### DIFF
--- a/acctz/acctz.proto
+++ b/acctz/acctz.proto
@@ -127,8 +127,10 @@ message UserDetail {
   // accounting a login that terminated before a username was received.
   string identity = 1;
 
-  // Privilege level configured on the system.
-  uint32 privilege_level = 2;
+  // user's privilege level, user class, user group, task group, etc.,
+  // which is implementation dependent.  Might be omitted if
+  // authentication fails.
+  string group = 2;
 }
 
 // Details of authentication
--- a/acctz/acctz.proto
+++ b/acctz/acctz.proto
@@ -94,28 +94,88 @@ message SessionInfo {
   // - multiple sessions over a single SSH connection
   // - multiple channels over a single gRPC connection
   string channel_id = 6;
+
+  // optional name of the tty, eg: console0, aux0, vty0, pty0.  If the
+  // connection is a serial line, such as a console, the adddress,
+  // port, and protocol fields above will be omitted.
+  string tty = 7;
+
+  // session status
+  enum SessionStatus {
+    SESSION_STATUS_UNSPECIFIED = 0;
+    SESSION_STATUS_LOGIN = 1;	// "start"
+    SESSION_STATUS_LOGOUT = 2;	// "stop"
+    SESSION_STATUS_ONCE = 3;	// simultaneous login, cmd, logout
+    SESSION_STATUS_ENABLE = 4;	// change privilege level
+    SESSION_STATUS_IDLE = 5;	// "watchog"
+  }
+  SessionStatus status = 8;
+
+  // User details
+  UserDetail user = 9;
+
+  // Authentication details
+  AuthenDetail authen = 10;
 }
 
-// Details of authentication - for gRPC or shell/vendor-CLI.
-message AuthDetail {
+// Details of a user
+message UserDetail {
   // Identity string is used to identify the user that executed the
   // command. For instance, it could be the spiffe-id in the case of
   // gRPC or unix-style user-name in the case of shell/vendor-CLI.
+  // This might be omitted if the username is unknown, such as
+  // accounting a login that terminated before a username was received.
   string identity = 1;
 
   // Privilege level configured on the system.
   uint32 privilege_level = 2;
+}
+
+// Details of authentication
+// While some authentication mechanisms authenticate the client only
+// once, at connection time, such as mTLS and shell, others authenticate
+// the client for each service_request, such as gRPC w/o mTLS or HTTP.
+message AuthenDetail {
+  // authentication type
+  enum AuthenType {
+    AUTHEN_TYPE_UNSPECIFIED = 0;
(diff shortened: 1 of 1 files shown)


acctz/acctz.proto
```diff
@@ -94,28 +94,90 @@ message SessionInfo {
...
+  string tty = 7;
+
+  // session status
+  enum SessionStatus {
+    SESSION_STATUS_UNSPECIFIED = 0;
+    SESSION_STATUS_LOGIN = 1;	// "start"
+    SESSION_STATUS_LOGOUT = 2;	// "stop"
+    SESSION_STATUS_ONCE = 3;	// simultaneous login, cmd, logout
+    SESSION_STATUS_ENABLE = 4;	// change privilege level
+    SESSION_STATUS_IDLE = 5;	// "watchog"
```
morrowc, line 110: 'watchdog' ?

morrowc (approved):
minus what looks like a typo, ok :)

acctz/acctz.proto
haussli: This is what tacacs+ (rfc8907) calls this update message.  ie: "still logged-in".  There probably is a better term than "IDLE"; if you have a suggestion, please share/change it.

haussli: I think that we are still waiting for comments from baires on these changes.  I suppose that should timeout at some point.

acctz/acctz.proto
```diff
@@ -94,28 +94,90 @@ message SessionInfo {
...
+    AUTHEN_STATUS_ERROR = 3;
+  }
+  AuthenStatus status = 2;
+
+  // In case of STATUS_FAIL/_ERROR, detail of the cause
+  string cause = 3;
+}
+
+// Details of authorization - all service_requests
+message AuthorDetail {
```
nmahabaleshwar, line 168: nit: Does it make sense to replace `Author` with `Authz` which is the word used for authorization in gNSI ?

```
AuthzDetail
AuthzStatus
AUTHZ_STATUS_UNSPECIFIED
.
.
.
AuthzStatus
```

haussli: Hey, I do not know; personally, I do not like the 'z' thing.  eg: it is ambiguous w/o more context; is that authentication or authorization.  In this case, I believe that I just continued the name from the previous version; I did not change it.

But, whatever the majority thinks is best, works for me.  Hopefully, Eben and Morrow will comment on this.

nmahabaleshwar: Can we just use the entire word `Authorization` ? `Author` is a noun and there is ambiguity on what it represents in this scenario.

Also should we move this to SessionInfo where we also have the Authentication information ?

erahn: I want to note: AuthN is the common abbreviation for authenticatioN. AuthZ is the common abbreviation for authorizatioN. Author is pretty ambiguous and confused me when I read this.

morrowc: I don't see that there is consensus to leave it alone, I'd err on the side of "less ambiguity" personally.
So... instead of 'bring me a rock', please use either:
   * full word (authorization / authentication / accounting)
   * industry standard common abbreviations (authn, authz)
   
   either of those tacts is ok for me :)

haussli: I prefer not to be involved in naming things.  I would like erahn and nmahabaleshwar to create a separate merge request to make these changes.

acctz/acctz.proto
haussli: @earies Do you have comments on this?  The window is closing on November 2.

acctz/acctz.proto
nmahabaleshwar: We can raise a different PR about naming. How about the other comment regarding the SessionInfo having the Authorization information along with the Authentication information ?

haussli: To be sure that I understand, you are asking if AuthorDetail should be moved from RecordResponse to SessionInfo (and therefore RecordResponse.SessionInfo)?

Is authorization part of the session, or part of the operation (ie: CommandService, GrpcService)?  My perspective is the latter, but so as not to duplicate the field in each operation message, it could be in RecordResponse.  Maybe I am wrong.

Is there a process advantage for the device if SessionInfo does not change for AuthorDetail,  where it can often just copy SessionInfo for most accounting messages in a particular session?

erahn: - AuthDetail (authen) should be under SessionInfo, because that's going to be consistent throughout the session
- AuthorDetail (author) I feel should be under both grpc_service and cmd_service. It's really associated with what was ran so it should be colocated with it.

nmahabaleshwar: We can leave it in RecordResponse to avoid duplication or have it in both `grpc_service` and `cmd_service`. Thanks @erahn

haussli: Hey, @nmahabaleshwar @erahn, I am resolving this so that we can move forward.  Please open an issue or a PR to address whatever you would like to see changed.

acctz/acctz.proto
haussli: Hey @earies ,  I am resolving this so that we can move forward. Please open an issue or a PR to address whatever you would like to see changed.

acctz/acctz.proto
nmahabaleshwar: Thanks @haussli

haussli:
> - AuthorDetail (author) I feel should be under both grpc_service and cmd_service. It's really associated with what was ran so it should be colocated with it.

Those are good arguments.  Please open a PR or MR with that.

PR #123, opened 2023-08-21
Update README for recent changes to services.

haussli:


acctz/README.md
```diff
@@ -9,24 +9,23 @@ remote collection service, primarily over a gRPC transport connection.
 ## Method of Operation
 
 Accounting Records are available at a gNSI origin:
```
earies, line 11: Thx - missed these in the previous commit..... question on the use of the word `origin` here - what is that meant to entail exactly?

It seems like this might have been an initial intention to replicate the gNMI concept of `origin`?

Otherwise its really just the base package name for the HTTP path header/proto def not including the version

haussli: Not sure; I assumed it meant "the root of the accounting tree", but its about as clear and useful as the northbound/SEWbound terms.  feel free to submit a MR to change it.

morrowc: I think I was told, way back when, that having the various gNSI parts represented in the OC tree of things would include gNSI parts being under their own 'origin' (root? or the level just below the root of the OC tree?).

earies: Very well possible from an implementation perspective but this would need to be dissected and refactored looking at the schema augments in gNSI as they stand today.

Currently, these schemas are all state that are augmenting into a tree represented by the `openconfig` origin and if the intention is to "hide" these augments to that origin and only be accessible to another (e.g. `gnsi.acctz`) then we have to open up the discussion for what this means for subtrees of different origins, parent nodes accessibility, etc..

And then it seems its not a single origin for `gnsi` by this which then begs the question as to why and what benefit in reality vs. overcomplication

If it truly is a demarc of microservice state and it should be represented separately then I would lean towards an entirely different YANG modeled subtree under a common origin.  If you are looking to attach this to a specific grpc-server instance under the `openconfig` origin, then you could very well consider pathz to gate the permissions to the subtree and not overcomplicate the use of `origins` (where you would also have to account for more origins in your authorization rule for `gnmi.Path`)

_NOTE_: The OpenConfig grpc-server modeling in general needs some refactoring imo to consider a more structural way to represent per service config/state and attachment points from those service schemas (among various other gaps)

morrowc: The 'origin' bit in OC/yang is super opaque to me... so yea, if this doesn't make sense 'origin == top of a tree naming thing' then ok.

haussli: Please move this discussion to a new issue/problem report.  It seem irrelevant to this particular merge request.

morrowc: ok, I think all set here now.

morrowc (approved):

morrowc (approved):

morrowc (approved):

PR #125, opened 2023-08-23
Fix the canonical order of "config" statement

nmahabaleshwar:
Fixed the order of `config` statement according to RFC 6020, Section 12

### DIFF
--- a/credentialz/gnsi-credentialz.yang
+++ b/credentialz/gnsi-credentialz.yang
@@ -22,6 +22,12 @@ module gnsi-credentialz {
         "This module provides a data model for the metadata of SSH and console
          credentials installed on a networking device.";
 
+    revision 2023-08-18 {
+        description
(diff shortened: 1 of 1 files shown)


credentialz/gnsi-credentialz.yang
```diff
@@ -22,6 +22,12 @@ module gnsi-credentialz {
         "This module provides a data model for the metadata of SSH and console
          credentials installed on a networking device.";
 
+    revision 2023-08-18 {
```
robshakir, line 25: You can probably just use a patch revision here if you'd prefer, but this is also fine.

robshakir (approved):

credentialz/gnsi-credentialz.yang
robshakir: Sorry, nonblocking.

Issue #111, opened 2023-08-10
Certz does not have counters per-profile


haussli:
certz supports multiple ssl profiles (RotateCertificateRequest.ssl_profile_id).  It also maintains counters exposed as
augment /oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/oc-sys-grpc:state:

This structure does not appear to be per-ssl_profile_id.  So,
1) it is global, which is not logical given its fields, or
2) it needs a ssl_profile_id field and a list keyed by that id, in yang parlance.

2 seems to be the answer, but I'm not sure what implementation to suggest here.

morrowc:
Sadly I keep avoiding learning much of anything about YANG.. but like a ton of other things on a network device where there are multiple instances of a thing, yea .. we want to count the relevant / critical bits related to those instances and not attempt to interpolate / infer what one instance of the thing is doing based on jiggery-pokery from remote.

so yea, please do 2 ... the path you point to doesn't look directly like it's 'ssl profile' related...

haussli:
I think that having an ssl-profile-id associated with each grpc-server would solve this.  Would you create a MR for  that?

PR #126, opened 2023-08-24
certz: add ssl-profile-id leaf

brianneville:
This leaf creates an association between a particular gRPC server and the ID of the SSL profile that it is running with.
see https://github.com/openconfig/gnsi/issues/111


haussli:
LGTM.

morrowc (approved):

PR #127, opened 2023-09-05
Remove not-supported deviations

nmahabaleshwar:
Removing the not-supported deviations for the below paths since they are required if gNSI is not configured.

```
/system/aaa/authentication/users/user/config/password
/system/aaa/authentication/users/user/config/password-hashed
/system/aaa/authentication/users/user/config/ssh-key
/system/aaa/authentication/users/user/state/ssh-key
```

If gNSI is enabled and its not desirable for these config/state to be accessed via gNMI, gNSI.pathz can be used to block access to these nodes.


tomek-US (commented):
As the author of the original version of this model I still believe it is important to keep the deviations. They are needed as they explicitly mark leaves that _cannot_ be supported if the gNSI is in use.

charanjith-anet:
> 

There can be some operators that would still like to use configure security using gNMI only. To support them, the vendor devices should continue to support the OpenConfig YANG paths. 

Do you have any suggestions on how to address this situation when these deviations are kept?

morrowc (approved):

morrowc:
seems fine, to me, in concept here, I think maybe the 2 folks (tomek, charanjith) have some issue to finalize conversaation on though.

charanjith-anet (approved):
Looks good

ncorran:
I dont know if @tomek-US is going to respond, but if the customer wants to use local config via gNMI then what about that being done via vendor native models?  It seems reasonable that openconfig is based on full use of the suite of gNxI protocols and gNSI supercedes these items in gNMI.

In particular gNSI clarifies that these items are not 'configuration' and are stored in a policy DB away from the config, so having gNMI put them into the running config adds inconsistency in the gNxI suite.

dplore:
These models will be removed from the gNSI repository.  This PR in the public openconfig repo replaces them (including removal of the deviations)

https://github.com/openconfig/public/pull/1037

dplore:
I will close this PR and introduce a new one to delete these models as superseded by https://github.com/openconfig/public/pull/1037)

PR #143, opened 2023-12-15
gnsi/pathz: clarify rotation behaviour for ongoing gNMI subscriptions

brianneville:
The index.md is updated with a section covering the expected behaviour for ongoing subscriptions when a server
experiences a pathz policy change.

This commit also updates the pathz.UploadResponse value so that
it contains information about ongoing subscriptions and whether
they would experience different behaviour under the incoming policy. .pb.go files were regenerated.

### DIFF
--- a/pathz/index.md
+++ b/pathz/index.md
@@ -93,7 +93,8 @@ sent, the proposed configuration is rolled back automatically.
 
 ### AuthorizationPolicy rotation and gNMI subscriptions
 
-When the pathz policy is rotated, ongoing gNMI subscriptions will be dropped.   
+When the pathz policy is rotated, ongoing gNMI subscriptions will be dropped.
+
 This forces gNMI clients to reconnect and ensures that all gNMI subscriptions
 are proceeding using the most recent pathz policy.
 
--- a/pathz/index.md
+++ b/pathz/index.md
@@ -91,6 +91,12 @@ out the action.
 If the stream is disconnected prior to the Finalize message being
 sent, the proposed configuration is rolled back automatically.
 
+### AuthorizationPolicy rotation and gNMI subscriptions
+
+When the pathz policy is rotated, ongoing gNMI subscriptions will be dropped.   
+This forces gNMI clients to reconnect and ensures that all gNMI subscriptions
+are proceeding using the most recent pathz policy.
+
 ## Open Questions/Considerations
 
 None to date.


pathz/index.md
```diff
@@ -91,6 +91,69 @@ out the action.
 If the stream is disconnected prior to the Finalize message being
 sent, the proposed configuration is rolled back automatically.
 
+### AuthorizationPolicy rotation and gNMI subscriptions
+
+When a gNMI client connects and begins a subscription, the active
```
ncorran, line 96: This PR changes the behavior of gNMI to make the pathz setup explicitly per subscription.

In some implementations (I would have guessed the majority) the pathz setup is a global policy, not per subscription and none of the PR applies to that case - i.e. on such a device all subscriptions would have their policy updated when the rotate is finalized.

I think this PR doesn't work well for all implementations as-is and should be changed to be more flexible.

What is the actual advantage of this setup, vs a central policy?  If the device has 100s of sessions/subscriptions this PR suggests it could have 100s of copies of a policy database (and that policy database could be huge itself - depending on the granulaity of the rules and the number of users/groups etc), which is a memory burden that seems inappropriate to push onto the device (and no obvious advantage - to leave subscriptions running against stale policy, getting access to something that has been disabled via a pathz update)

At the very least I think there should be some device capability so the device can notify if it has per session or global pathz policy, and then only in the cases where there is per session policy would these extra interactions apply.

brianneville: >This PR changes the behavior of gNMI to make the pathz setup explicitly per subscription.
>
>In some implementations (I would have guessed the majority) the pathz setup is a global policy, not per subscription and none of the PR applies to that case 

I would say that it clarifies the behaviour, rather than changes it ;)
There isn't currently anything in the spec which describes how pathz rotation should affect ongoing subscriptions. There's also no coordination between vendors on this - if some implementations happen to have settled on a single global policy which operates in the way you describe, then this is pure coincidence. 

>What is the actual advantage of this setup, vs a central policy?

I'll address the limitations/issues of these two approaches that I can think of, and then explain why i think the approach in this PR is the most appropriate.

### Approach 1: a single pathz policy is applied to all subscriptions and updates as policy rotates (this is the central policy you describe):
The issue with this twofold, and comes from two main considerations about the behaviour of the server as the policy would change.
Firstly, consider that that the policy contents are actually relevant throughout the life of the subscription, as new items are populated into the YANG tree.
e.g. imagine we have a subscription to a list `/a/b` and a new list item `/a/b[k0=X]` appears in the YANG. At this point presumably the policy is now consulted to determine whether or not we can send this update over the subscription (im making this assumption based on the fact that you're trying to minimise memory, by not snapshotting the policy at the beginning of the subscription - correct me if this is not the approach that you're describing).

Secondly, consider that gNMI doesn't have any mechanisms for updating the gNMI client that their pathz policy has been updated. We also shouldn't build this kind of mechanism into gNMI. Pathz should remain a transparent/invisible authorization layer which is only relevant to the the gNSI client/security team, and not something which every gNMI client needs awareness of.

With those two points, consider a subscription from the perspsective of a client who is having the pathz policy updated in the background like you describe.
- A gNMI client with username "bob" subscribes to list `/a/b`
- A new list item appears at `/a/b[k0=X]`, and under the current policy, the update gets to go through to the client
- a gNSI client rotates a new pathz policy, denying the user "bob" access to `/a/b[k0=X]`

At this point, we're caught in a strange state.
the item `/a/b[k0=X]` will continue to update (or even be deleted), but "bob" wont see any of those updates/deletion. 
However "bob" hasn't recieved any new data for this leaf, and there's no signal in the gNMI subscription to indicate that "hey, your pathz policy was updated and this is where you stop getting updates". From bob's perspective, he is oblivious to this, and may continue operating on the assumption that the outdated value is still current.

Another similar issue occurs where a policy is made less restrictive.
For example:
- A gNMI client with username "bob" subscribes to list `/a/b`
- the current policy dictates bob has only access to `/a/b[k0=X]`, so he receives only updates for that item.
- a gNSI client rotates a new pathz policy, allowing the user "bob" access to `/a/b[k0=*]`

Again, we're caught in a strange state here
Bob suddenly starts receiving updates for items which under the initial sync, he has no previous awareness of.
For example, he may receive a delete for the list item `/a/b[k0=Y]` which he may have never received updates for.

So the problem with approach 1 is that after every update, we risk leaving clients in these strange states, and the gNMI subscription becomes unpredictable from their perspective.
I dont imagine any network operator would want to experience a subscription which was performing like this.
Receiving gNMI updates from a path and then suddenly not receiving them would be seen to imply something about the state of the system (e.g. if a client sees interface counters stop incrementing, have we stopped traffic? or is it because the pathz policy has been rotated?). 
Likewise, receiving information like deletions about previously unknown elements could be mistaken as a bug in the client, if the client doesn't know that the pathz rotation occured. 
Even if it was not seen as a bug, it seems like an incredibly bizarre case which pathz is now forcing every gNMI client to build handling for.
As you point out we can have lots of gNMI sessions open, and with more sessions and arbitrary rule updates, the likelihood of a network operator having a client left in these strange states increases.

### Approach 2: snapshot policy and leave subscriptions open using it (i.e. the approach described in this PR)

This has a few limitations:
There is unavoidable memory overhead for the server.
How much memory is required is a implementation detail, but yes this approach does unavoidably open the door for more memory consumption on the server-side as you noted.
There is also the risk that we are leaving open subscriptions which were running on a stale policy, as you pointed out.

----

So coming back to your question:
> What is the actual advantage of this setup, vs a central policy?

The advantage of this setup is that we avoid confusing clients with the issues from approach 1 described above.

The approach in this PR is the only way I can think of to guarantee stable gNMI subscriptions, while a central policy leads to complex client/server interactions since the policy is changing in the background and the view of the YANG is unpredictable. Without total awareness of the gNSI pathz policy state, a gNMI client can't tell whether a subscription is evolving in a particular way due to the underlying system state, or due to the pathz policy being rotated over the course of the subscription.
Yes, we're trading off some extra memory consumption for this stable client behaviour, but this is a worthwhile trade in my eyes. 

The need for stable client behaviour is a main driving force behind this PR. 
However, regarding stale subscriptions, the modified `UploadResponse` in this PR allows for the gNSI client to be aware of which gNMI clients were subscribed, as well as whether this access would now be impacted by the incoming policy. 
This can be used to combat stale subscriptions, as network operators are now alerted to the state of the ongoing subscriptions and whether or not the incoming policy would modify their access. The network operator can then disconnect whatever subset of clients it wants, based on its own idea of which paths are sensitive (e.g maybe its critical that clients should be disconnected if the incoming policy changes their access to `/a/b`, but clients subscribed to `/x/y` can remain on the old policy, even if the incoming policy would give them slightly different access).

>At the very least I think there should be some device capability so the device can notify if it has per session or global pathz policy, and then only in the cases where there is per session policy would these extra interactions apply.

I think this could complicate things, as it would mean that network operators who use products from multiple vendors now could have to handle the limitations of a central policy anyway. 
Having parts of a network operating with a fundamentally different approach to pathz would get confusing to deal with.
I'll leave it to the repo maintainers (@morrowc @marcushines) to comment on whether or not the spec should have something like this, but in my opinion it would be a step in the wrong direction.

marcushines: +1 to Brian's comments.  They distill the discussion on this proposal over the last week

> At the very least I think there should be some device capability so the device can notify if it has per session or global pathz policy, and then only in the cases where there is per session policy would these extra interactions apply.

I think having two different server side implementations would be problematic in having clients understand the expected behavior.  Also I think the approach presented here better aligns with Authz for consistency

We would not expect Authz policy changes to "interrupt" any RPC in progress on the system. This would keep Pathz operating the same manner.

marcushines (approved):

pathz/index.md
ncorran: I agree that for approach 1 we have a strange situation if pathz changes and the subscription stays open (but impacted, potentially losing access to data it has previous seen).  I think the solution to this is for the server to close all open subscriptions when the policy is changed (perhaps optimized - device choice - to only close subscriptions where the pathz change will impact the policy)

I still don't agree with the idea of forcing the server to keep a per-session/subscription policy database.  The policy is in my view a global entity (with potentially significant cost to duplicate).  We would not want to create many instances.

I think it is reasonable for the server to EITHER support
- multi instance policy, keep old subscriptions open AND provide and api for the user to check if the susbcription has been impacted by a pathz policy change, so the user can restart the subscription to get the latest policy applied (or know they are stale)
OR
- the server auto detects and closes (forcing a client reconnect and resubscribe w latest pathz), so the server can keep a global pathz policy

Personally I think the second approach is the better approach to always take and would remove the need for the complexity (pushed onto the device) in this proposed set of changes.  IMO the purpose of a pathz change is to enforce a policy on all clients, NOT to allow clients to bypass the pathz policy changes by keeping an old session open as long as possible.  Potentially sensitive information could be leaked with this 'delayed' approach (e.g. consider I apply the new policy, then configure new sensitive info assuming the policy is in force and long-standing clients can receive this data via old subscriptions... it leaves the pathz policy impossible to enforce when it matters).

brianneville: From a vendors perspective I can see why having the server disconnect clients is appealing, however from a network operator's perspective, I don't think it is as appealing. 

The main issue with the server closing subscriptions is that it would force complexity onto all network operators.
Consider the following:
As the gNMI subscribers do not have any awareness of pathz, there is not any way that the server can signal them that this disconnect was triggered by pathz and that they should reconnect. 
This means that as far as I can tell, all network operators will have to do one of the following:
- build retry logic into all subscriptions, where this logic will continue to blindly retry subscriptions without any way to distinguish pathz-related disconnects from any other kind of disconnect.
- build some coordination layer which _all_ gNMI clients have to incorporate, where this coordination layer informs gNMI clients that a pathz rotation has occured and that they should reconnect and/or gives clients advance warning in case they want to gracefully disconnect. This coordination layer would be driven by the gNSI client / security team.

On a more philosophical note:
For the other elements of the Authentication/Authorization stack we can see that security policy/configuration being something which is only consulted once, at connection time. For example, the gNSI Authz policy is consulted atomically at connection time, at which point the connection is rejected or accepted. Connections/RPCs which are rejected by the current Authz policy are not then held in any kind of purgatory until a new policy is rotated which would accept them, and similarly RPCs which are accepted won't be disconnected mid-RPC if the policy changes.

Likewise for authentication, TLS/mTLS handshakes only occur once as the connection is being established, at which point the user is taken to be authenticated for whatever operation they're running, regardless of if those credentials are updated in the background after the connection is set up.
So the proposal in this PR (multi instance policy, keep old subscriptions open) would just be aligning pathz with this existing paradigm.

>IMO the purpose of a pathz change is to enforce a policy on all clients, NOT to allow clients to bypass the pathz policy changes by keeping an old session open as long as possible 

I think it depends on how you view the policy really.
I was/am viewing it as something we lookup once which tells us how to proceed and filter the subscription, rather than something which we're continually re-evaluating the client against.
Point taken about having old subscriptions still active after rotation though - the example you gave could be a valid concern under the current proposal.
I'll let the gNSI maintainers comment here on how they envision this policy as applying to the subscription (as something to continually re-evaluate or as something which should be consulted once to set up filtering)

ncorran: The server can always choose to close a subscription (e.g. low system resources, or any other internal reason).  There is no gnmi message to pass a 'reason' so the client has to just cope with this, and really the only option is to try to reconnect/resubscribe.  When that happens an error condition can be returned if the connection is rejected (e.g. too many subscriptions or not enough memory or whatever).  For the pathz change case, the resubscribe would be expected to work (and resync and all is well).

There is no complexity added by a close in this case, as it can already happen at any point - and so a reconnect scheme is already needed by any users/controller system.

I'm not suggesting any purgatory here - at the subscribe time it will succeed or fail based on validity+pathz, and pathz approve/deny (of the data that can be returned) is constant for the duration of the subscription - this is the same with both approaches.

I'm just removing the complexity of messages to and fro about pathz changes and just closing a subscription when it would otherwise be impacted (respecting the security provided by pathz always).

brianneville: I've updated the PR adding a new field into the `FinalizeRequest` message which the client can use to determine how the server should deal with subscriptions, please take a look.

The workflow you're describing would then be enabled by providing the `FINALIZE_EFFECT_CLOSE_IMPACTED_SUBSCRIPTIONS` enum as the `FinalizeRequest.effect` field.

ncorran: Just to note that the whole point of the original comments is that the device may not support multiple options, but rather work in a specific way - which the client cannot control.  I'm not sure field meets the requirement.

ncorran (changes requested):
The PR still doesn't account properly for devices that have a single central policy DB only, not (up to) one per subscriber... and/or devices that want to enforce a stricter security posture.

marcushines:
I think at this point to resolve the issue we should look at simply dropping the connections on rotation and just forcing the new policy to be applied (this really only applies to streaming methods)

brianneville:
> I think at this point to resolve the issue we should look at simply dropping the connections on rotation and just forcing the new policy to be applied (this really only applies to streaming methods)

Sounds fine - I've update the PR so that this approach is referenced in the index.md, please take a look @marcushines @ncorran

ncorran (approved):
Approved from Cisco

PR #147, opened 2024-01-26
Add name of the component name where the accounting response originated

nmahabaleshwar:


morrowc (approved):

LimeHat:
Is there an example how is this field supposed to be used? I.e. what is the "component" in the context of this service?
@nmahabaleshwar @morrowc

nmahabaleshwar:
> Is there an example how is this field supposed to be used? I.e. what is the "component" in the context of this service? @nmahabaleshwar @morrowc

The "component" is the physical entity where the accounting event occurred like a supervisor, line card etc. The vendor specific entity name is populated in this field.

Issue #137, opened 2023-12-01
Clarity needed (in a comment) for acctz RecordRequest keepalives


haussli:
Since the keepalives are not ACKs of any sort, clarity should be added about the timestamp in the KA RecordRequest.
What should it be set to?
Should the server ignore it?  Else what is the behavior?  If it is ignored, how does a client stop the stream and restart it from a particular timestamp.
Also reference issues/77

morrowc:
Oh, this is a good point. I believe in initial thoughts about this we had defined a keep-alive proto/packet... but were dissuaded from that path for reasons which aren't in my memory banks today :(  I think minus the keepalive officially being a packet/message, you can certainly just heartbeat requests on a reasonable timer.

It seems likely that in some deployments you may go quite some time between 'changes' on devices, so your heartbeats may just not get returned data for long intervals of time. Decisions about: "is this connection dead? Is the far end overworked and can't reply, but still there??"  seem like they get fixed at the TCP layer, mostly. It would be unfortunate if the client here decided after N no-replies to bounce the session, that is unnecessary work on the overall system, and possibly a confusing behavior :(

I think some thought/guidance on the system here is in order... I suspect that at least the idea that:
  0) start session, ask for <client configured> history be requested ("hi, give me everything since 27 days ago!")
  1) delivery from server -> client of the backlog avaialble
  2) sending of 'record request' with timestamp=record-received-timestamp
  3) process replies if any are received
  4) goto 2

is a good starting point, for me anyway, in the discussion.

nmahabaleshwar:
I think in the current proposal there is no difference between a keep-alive and request message asking for records. Atleast the server treats every `RecordRequest` message as `The client is asking something, let me check if I have something to be sent`, it sends RecordResponses if there is any new event which happened after the timestamp in the message. It also uses this message as an indication that the client is still alive and consuming records. As @morrowc mentioned, some of the concerns about the liveness of the far-end can be handled by TCP so I am not sure if we need an application level keep-alive mechanism.

morrowc:
It's almost certainly the case that I imagined this service in a totally incorrect manner :(
(this is probably where my original keep-alive type handling was supposed to be happening... that's since morphed)

Ok, so originally I viewed this as:
  1) client connects to the server/device and asks for accounting records from at least time-X.
  2) server/device sends all records in bunch (how that happens on the wire isn't important to me).
  3) server/device keeps on sending new logs as they are created
[...]

PR #148, opened 2024-01-30
Update docs & comments for the behavior of the RecordSubscribe service,

haussli:
removing the keepalive concept and mentioning the behavior the server can adopt to timeout a dead client.

Addresses issue #137

### DIFF
--- a/acctz/acctz.proto
+++ b/acctz/acctz.proto
@@ -66,10 +66,13 @@ service Acctz {
   // System should be sent to the Collector. A current time indicates that only
   // new messages should be sent.
   //
-  // The Collector must send a RecordRequest at least every 60 seconds, in
-  // order to signal its liveliness to the system. Failure to send a
-  // RecordRequest for more than 120 seconds will cause the connection to be
-  // reset.
+  // The stream continues ad infinitum, until the gNSI session is severed.
+  // The gNSI TCP connection could timeout due to expiration of the TCP
+  // keepalive mechanism, or the server could disconnect the client if the
+  // output buffer to the client remains full for a long timeout period,
+  // implying a stuck client (see also getsockopt(SIOCOUTQ) and
+  // https://datatracker.ietf.org/doc/draft-ietf-idr-bgp-sendholdtimer/).
+  //
   rpc RecordSubscribe(stream RecordRequest) returns (stream RecordResponse);
 }
 
--- a/acctz/README.md
+++ b/acctz/README.md
@@ -12,7 +12,7 @@ Accounting Records are available at a gNSI origin:
    gnsi.acctz
 
 Records may be streamed from a system at request of the remote collector,
-via the AccountingPull() service/rpc.
+via the RecordSubscribe() service/rpc.
 
 Configuration of the Accounting service is made through standard
 gNxI methods using the defined YANG model.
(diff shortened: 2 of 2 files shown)


morrowc (approved):

acctz/acctz.proto
```diff
@@ -66,10 +66,13 @@ service Acctz {
   // System should be sent to the Collector. A current time indicates that only
   // new messages should be sent.
   //
-  // The Collector must send a RecordRequest at least every 60 seconds, in
-  // order to signal its liveliness to the system. Failure to send a
-  // RecordRequest for more than 120 seconds will cause the connection to be
-  // reset.
+  // The stream continues ad infinitum, until the gNSI session is severed.
```
nmahabaleshwar, line 69: Can we also update the comment at line 56 ?

```
If this results in no records to send, the server should return an OK error and zero records.
```

I don't think an OK error should be sent to the collector if no records remain to sent, the connection stays open and RecordResponses will be sent for future events.

haussli: What should it return?

If the subscription succeeds, but either the history is empty or a time is requested that both does not correspond with a record in the history and there are no records after that time in the history, then the current text seems reasonable.  For example, the client might request now() or a future time (T+60s, which the server should probably be permitted to assume to be now()).

nmahabaleshwar: If we return an OK error in those scenarios, the stream will be closed which we don't want right ?

In all those scenarios, we want the stream to be open and the collector should get RecordResponses when accounting events occur after the given timestamp in the original RecordRequest message.

If the server wants to explicitly signal the client that `I don't have anything for you now, but will send records in the future`, maybe we create a new message or add a new field in the RecordResponse which indicates that.

morrowc: Ah, perhaps: "if there's an error on the RPC the connection is closed"

is the missing part here in the conversation?
Generally, Error on a streaming RPC means: "oops, bad things, please restart your process"
so we should say that no error is sent, but that also no records are sent (because there are none).

Perhaps REALLY in the text about the startup  the reply is something like:

  "device returns as many records as will satisfy the request, or none if there are no records which satisfy the request. The connection remains open such that future records may be streamed to the client as they are available."
  
Or something along those lines, at any rate.

haussli: Agreed, except, doesn't it also return a result code in both cases?

morrowc: that's a fair point, we COULD have the server send back a status: ok  (this is nirajan's text mostly I think)
but really .... we connect, we send a 'hope you have mail for me!, but I'm hanging out here anyway for later.'
and that's ok. Whether we get 0, 1, tons of records on our first connect is a little unimportant.

There IS a 'did you hear me???' problem, but that might be solve either in the metrics/alerting pipeline ("that device is oddly silent on acctz.. whats up?") or with a simple: "if nothing heard in N mins, disconnect and reconnect"

that model (disco/reco) is ALSO problematic, because it doesn't really solve anything and never surfaces the 'that device is oddly silent...' to the alerting pipeline.

morrowc: oops... and.. nirajan, if you think there's wording fixes / etc please don't hesitate to send along :)

haussli: Ja, thanks for the explanation, morrow.  Please suggest the text nirajan, or suggestion for the text to consider is here: https://github.com/haussli/gnsi/commit/33b831a713f8b5e41495d713414d193068701689

nmahabaleshwar: Raised https://github.com/openconfig/gnsi/pull/151 for this

haussli: Thanks!

PR #149, opened 2024-01-30
Argument to RecordSubscribe should not be a stream.

haussli:
Missed this change in the previous merge.

issue #137


morrowc (approved):
Would ask a niranjan to have a gander here, but... this SEEMS ok to me:
  1) the request only has a timestamp: "tell me anything since X"
  2) I don't see you wanting to make that request often on the same connection
       (you are just waiting around for more data, no need to re-ask)
       
overall seems ok, but :) 2x check would be nice.

morrowc:
huh, I can't add niranjan... maybe we poke him in another method to have a look

nmahabaleshwar:
> Would ask a niranjan to have a gander here, but... this SEEMS ok to me:
> 
> 1. the request only has a timestamp: "tell me anything since X"
> 2. I don't see you wanting to make that request often on the same connection
>    (you are just waiting around for more data, no need to re-ask)
> 
> overall seems ok, but :) 2x check would be nice.

This looks fine, lets incorporate the text in https://github.com/haussli/gnsi/commit/33b831a713f8b5e41495d713414d193068701689

nmahabaleshwar:
> > Would ask a niranjan to have a gander here, but... this SEEMS ok to me:
> > 
> > 1. the request only has a timestamp: "tell me anything since X"
> > 2. I don't see you wanting to make that request often on the same connection
> >    (you are just waiting around for more data, no need to re-ask)
> > 
> > overall seems ok, but :) 2x check would be nice.
> 
> This looks fine, lets incorporate the text in [haussli@33b831a](https://github.com/haussli/gnsi/commit/33b831a713f8b5e41495d713414d193068701689)

Raised https://github.com/openconfig/gnsi/pull/151 for this

PR #162, opened 2024-02-23
Provide clarification on the field component_name

nmahabaleshwar:
Provide clarification on the field `component_name`.

### DIFF
--- a/acctz/acctz.proto
+++ b/acctz/acctz.proto
@@ -290,7 +290,14 @@ message RecordResponse {
   // accomplish the request on the system.
   repeated string task_ids = 7;
 
-  // Name of the component where the accounting record originated from
+  // Name of the component in a vendor system configuration where the accounting
+  // record originated from. This helps to indentify the source of the accouting
+  // record when multiple components can generate them. This is how the
+  // component is identified in the components hierarchy in the
+  // openconfig-platform model.
+  // For reference: /components/component[name="chassis0"]
+  // Eg: component_name = "linecard0"
+  //     component_name = "chassis0"
   string component_name = 8;
 }
 


morrowc (approved):
Might wait on heas@ to have a look at this.

haussli:
LGTM.  I would add another eg for clarity; eg: component_name: chassis0/linecard0

nmahabaleshwar:
> LGTM. I would add another eg for clarity; eg: component_name: chassis0/linecard0

Added another example as suggested

LimeHat:
Can you elaborate a bit on how it is supposed to be used / what is the purpose of the field? In what cases, for example, linecard0 component is used?

I would imagine that in most cases, accounting records are generated by control modules (in chassis-based systems) or a control module equivalent in a fixed system. Does it mean that the chassis returns the name of the active control module (at the time when the record was created)?

LimeHat:
Yeah, it is more or less clear with the control modules. 
What would constitute an accounting event on a line card? Is it for systems that allow users to directly connect to each line card (e.g. the linecard has a management IP, etc)?

haussli:
Some systems have redundant routing engines, multiple chassis with their own routing engines/control cards, and some have a linecards/cards in general that have a cli accessible in-line or from a serial port.  It is reasonable to expect that these cards might log/account these activites.  This field allows the origin or the records to be identified, which I would expect to be platform/vendor specific.  eg: (WAGs)
fpc0
chassis1/fpc0
re0
re1
chassis1/re1
0/RP0/CPU0

acctz/acctz.proto
```diff
@@ -290,7 +290,11 @@ message RecordResponse {
   // accomplish the request on the system.
   repeated string task_ids = 7;
 
-  // Name of the component where the accounting record originated from
+  // Name of the component in a vendor system configuration where the accounting
+  // record originated from. This helps to indentify the source of the accouting
+  // record when multiple components can generate them.
+  // Eg: component_name = "linecard0"
+  //     component_name = "chassis0/linecard0"
```
marcushines, line 297: This really should be the gnmi.Path string representation of the component

nmahabaleshwar: Updated the description

ncorran (approved):

morrowc (approved):

PR #168, opened 2024-03-05
Add ExistingEntity to allow copying Entities from existing profiles.

marcushines:
this allows new profiles to utilize either system default or other profiles Entities such as trust bundle, CRLs, Auth policy, or certs.


certz/README.md
```diff
@@ -106,6 +106,9 @@ The system will start with this profile and either bootz or enrollz will be resp
 
 Call `Certz.AddProfile` RPC with the `ssl_profile_id` field specifying the ID
 of the new SSL profile.
+If the new profile wants to use existing artifacts from other profiles, this can accomplished via sending `Entity` messages with
```
morrowc, line 109: this can BE accomplished

morrowc, line 109: I'd also argue that 'wants' in the first part probably should be something else.. for instance:
   "If the new profile should use existing...." ?
   
 (not a strong preference I guess on rewording)

morrowc (approved):

morrowc (approved):

PR #169, opened 2024-03-06
acctz: update GrpcService payload field

brianneville:
Udpate the payload field to be a string value.
This is used instead of google.protobuf.Any as
google.protobuf.Any cannot be unmarshalled into
a known proto message type if it is truncated
(as google.protobuf.Any is encoded).

Also add a comment specifying that this value
may be partially redacted in the case where we
want to avoid leaking sensitive information from
the device (such as private keys rotated via
gNSI Certz or gNSI Credentialz).


acctz/acctz.proto
```diff
@@ -284,14 +284,24 @@ message GrpcService {
...
   bool payload_istruncated = 5;
 
   // Authorization details
   AuthzDetail authz = 6;
+
+  // A string describing the payload of the RPC.
+  // This may be truncated.
+  // It also may be partially redacted to avoid leaking
+  // sensitive info (e.g. private keys rotated via certz/credentialz)
+  string payload = 7;
```
marcushines, line 304: in general i think the proto any is nice as it allows for parsing payload for rpc services or other services which can provide it 
Agree that logging this representation will truncate the bytes and cause an error - if you "stringified" the proto before logging though it would "mainly" readable

marcushines: i would see us changing this to be a oneof though to allow a string type for "simple logging" then a proto any for "structured logging"

brianneville: ack - made this be a oneof in the latest commit to allow either `string` or `google.protobuf.Any` to be sent

acctz/acctz.proto
```diff
@@ -284,14 +284,32 @@ message GrpcService {
...
   AuthzDetail authz = 6;
+
+  // the payload of the RPC, as either a string or proto-encoded.
+  // Note that this may be partially redacted to avoid leaking
+  // sensitive info (e.g. private keys rotated via certz/credentialz)
+  oneof payload {
+    // the payload in proto form.
+    // (suitable for small messages which don't require truncation)
+    // This can be unmarshalled by the acctz client.
+    google.protobuf.Any payload_pb = 7;
```
marcushines, line 307: proto_val

string_val is more aligned with how we have done these before

brianneville: updated in most recent commit

marcushines (approved):

PR #171, opened 2024-03-08
Add detailed scenarios on when history_istruncated is set to True

nmahabaleshwar:
`history_istruncated` in a RecordResponse is set to True if the gnsi.acctz server senses that there might be some events which have occurred prior to the timestamp in the RecordResponse message for which it might be missing RecordResponses. This helps the gnsi.acctz client to stitch a timeline where it can detect potential gaps in the timeline.


haussli (commented):
LGTM

morrowc (approved):

PR #172, opened 2024-03-13
certz: support specifying cert/key by enum

brianneville:
use oneofs and enums to specify the source
of the certificate and private keys for keys/certs which are already present on the device.
In particular this is relevant in the case of attestz, where OIDevID and IDevID certs and keys may be
stored in the TPM, see
https://github.com/openconfig/attestz/pull/36/


certz/certz.proto
```diff
@@ -453,11 +453,52 @@ message Certificate {
...
+    CertSource cert_source = 6;
+  }
+
+  enum KeySource {
+    // KEY_SOURCE_UNSPECIFIED can be used in the case where the
+    // certificate does not need a private key (e.g. trusted certs)
+    KEY_SOURCE_UNSPECIFIED = 0;
+
+    // keys stored in TPM
+    KEY_SOURCE_OIDEVID_TPM = 1;
```
marcushines, line 488: the oidevid key is the idevid key so shouldn't need this one

certz/certz.proto
```diff
@@ -453,11 +453,52 @@ message Certificate {
...
   // Not needed if the device knows the private key or does not have to know it.
-  bytes private_key = 4;
+  // Deprecated in favour of private_key_type
+  bytes private_key = 4 [deprecated=true];
+
+  enum CertSource {
+    CERT_SOURCE_UNSPECIFIED = 0;
+
+    // certificates stored in TPM
+    CERT_SOURCE_OIDEVID_TPM = 1;
```
marcushines, line 468: the certs don't have to be stored in tpm so i would suggest just 

CERT_SOURCE_<NAME>

certz/certz.proto
brianneville: removed in latest patch

certz/certz.proto
brianneville: ack, removed `_TPM` suffix from `CERT_SOURCE_OIDEVID_TPM` and `CERT_SOURCE_IDEVID_TPM` in latest patch

marcushines (approved):

certz/certz.proto
```diff
@@ -453,11 +453,51 @@ message Certificate {
...
+
+    // key is IDevID key stored in TPM
+    KEY_SOURCE_IDEVID_TPM = 1;
+
+    // key was generated with a CSR as part of rotation
+    KEY_SOURCE_GENERATED = 2;
+  }
+
+  oneof private_key_type {
+    bytes raw_private_key = 7;
```
jenia-grunin, line 495: Consider comment on the format of the private key: e.g. PEM-encoded.

certz/certz.proto
```diff
@@ -453,11 +453,51 @@ message Certificate {
...
+  enum KeySource {
+    // KEY_SOURCE_UNSPECIFIED can be used in the case where the
+    // certificate does not need a private key (e.g. trusted certs)
+    KEY_SOURCE_UNSPECIFIED = 0;
+
+    // key is IDevID key stored in TPM
+    KEY_SOURCE_IDEVID_TPM = 1;
+
+    // key was generated with a CSR as part of rotation
+    KEY_SOURCE_GENERATED = 2;
```
jenia-grunin, line 491: Will the device know how to find the matching generated private key?

jenia-grunin (approved):

certz/certz.proto
brianneville: done in latest patch - these keys will be PEM encoded

certz/certz.proto
brianneville: In the latest patch I updated this comment to say 
```
// key was generated with a CSR as part of the current Rotate RPC
```
So that there is no ambiguity about which rotation we should be considering here. 

The device should know how to find this private key, it can just retain a reference to it over the duration of the ongoing Rotate RPC.

marcushines (approved):

PR #175, opened 2024-03-14
acctz timeout behavior was removed; remove it from the yang model.

haussli:


nmahabaleshwar:
The YANG files are moved to the OpenConfig repo. We need to fix it https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-acctz.yang#L132

morrowc (approved):

morrowc:
> The YANG files are moved to the OpenConfig repo. We need to fix it https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-acctz.yang#L132

oops, we can move this in a new PR I believe.

PR #186, opened 2024-04-24
Fix typos in acctz.proto

gmacf:


morrowc (approved):
wow, thanks!
(for as often as I complain about folks typing skills.. I should remember I also need help :) )

PR #188, opened 2024-04-30
readd _grpc.pb.go files by changing the proto compiler from the default bazel io provides

marcushines:


dplore (approved):

WORKSPACE
```diff
@@ -51,7 +51,7 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 
 go_rules_dependencies()
 
-go_register_toolchains(go_version = "1.18")
+go_register_toolchains(go_version = "1.20")
```
bstoll, line 54: Should this be 1.22 and in agreement with go.mod?

bstoll (approved):

PR #197, opened 2024-07-02
Fix typo in `certz.proto`.

robshakir:
```
* (M) certz/certz.proto
   - trust_bundle_pcks7 --> trust_bundle_pkcs7 in the entity oneof.
```

This looks like a typo -- not sure if this field has been used, if so
we should probably either mark this as a known error and continue along
our merry way, or determine whether we think there are any textprotos on
disk anywhere that would mean that we need to mark field 8 as deprecated.

### DIFF
--- a/certz/certz.proto
+++ b/certz/certz.proto
@@ -564,7 +564,7 @@ message Entity {
     CertificateRevocationListBundle certificate_revocation_list_bundle = 5;
     AuthenticationPolicy authentication_policy = 6;
     ExistingEntity existing_entity = 7;
-    TrustBundle trust_bundle_pcks7 = 8;
+    TrustBundle trust_bundle_pkcs7 = 8;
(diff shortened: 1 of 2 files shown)


certz/certz.proto
```diff
@@ -564,7 +564,7 @@ message Entity {
     CertificateRevocationListBundle certificate_revocation_list_bundle = 5;
     AuthenticationPolicy authentication_policy = 6;
     ExistingEntity existing_entity = 7;
-    TrustBundle trust_bundle_pcks7 = 8;
+    TrustBundle trust_bundle_pkcs7 = 8;
```
marcushines, line 567: add new tag and deprecate old one

robshakir: Done.

marcushines (approved):

PR #200, opened 2024-07-24
Add a `profile` field to the `authz.Rotate` RPC.

robshakir:
```
 * (M) authz/authz.proto
  - Where a gNSI server supporting the authz service is running on
    a target that supports multiple downstream gRPC services, it is
    currently required that the same authz policy is applied to all
    of them. Particularly, if the downstream gRPC servers MAY expose
    the same service, it is not possible to have different policies
    for each server. This PR adds a `profile` field that allows
    such endpoints to be disambiguated, similarly to the `profile`
    that is provided by certz.
```

We have a use case where a target exposes a single gNSI server instance for a
set of downstream binaries (in this case, containers running on a device). In
such cases, it is desirable that we can expose common services per container for
common operations, but would like to have distinct authz policies for each.
The `profile` field allows each container to have its own authz profile that is
mapped to by the gNSI.Authz server.


authz/authz.proto
```diff
@@ -148,6 +148,19 @@ message RotateAuthzRequest {
...
+  // gRPC endpoints run by the target is assumed. Where non-default policies
+  // are to supported by an endpoint, the value of the profile determines which
+  // set of policies are to be rotated.
+  //
+  // It is not permitted to rotate multiple profiles' policies at the same time
+  // - if a `Rotate` RPC is in progress for profile "foo" (i.e.,
+  // `upload_request` has been supplied, but the rotation has not been
+  // finalised), and a second `RotateAuthzRequest` with `upload_request` is sent
+  // from the client for profile "bar" this is considered an error.
+  string profile = 4;
```
marcushines, line 163: i think we need to think through how this really gets implemented on device based on previous tls_profile support

robshakir: I think you have two options (I assume you mean `ssl_profile_id`).

1. SSL profiles and authz profiles are independent -- daemons can be configured to use a non-default SSL profile ID, and a non-default authz profile ID independently. In this case, similarly to certz, the gRPC server watches a specific filesystem path for both the SSL cert, and a potentially different one for the authz policy.

2. SSL profiles and authz profiles are tightly coupled -- daemons have a single config that indicates their `ssl_profile_id` which indicates their authz `profile` name too. In this case, the daemon still has two watch two paths, they are just tightly coupled to one another.

As per chat elsewhere, our policy here is inbound -- so is based only on the client's SPIFFE ID. Given that this is the case, there shouldn't be a hard dependency on the server side on its own ID -- thus, it's possible to `ssl_profile_id = FOO` and authz `profile = BAR`.

If we do choose to tightly couple them, this will create an "odd" dependency between `certz.AddProfile` and `authz.Rotate` since we'll need to have done a `certz.AddProfile` to be able to do an `authz.Rotate` for a particular profile. This might be OK -- but it feels a bit strange (`AddProfile` would essentially be a top-level concept in this case rather than a `certz` one).

The alternative to the proposal herein is that we don't support `profile` in authz -- and rather say "you must run different `gnsi.Authz` endpoints if you want to support multiple gRPC servers that have different policies". This seems a bit suboptimal.

The other thing we could do -- if we thought that this was very niche, is pack the `profile` into metadata for the RPC, but this feels like hiding functionality and hurts testability IMHO.

robshakir: Clarified to make case 1 what is expected, and renamed the field `authz_profile_id` per discussion.

marcushines (approved):

Issue #205, opened 2024-09-19
Support for RSA 3072 and ECDSA_P_384


rohit21187:
Hi I see that in [gnsi/credentialz/credentialz.proto](https://github.com/openconfig/gnsi/blob/main/credentialz/credentialz.proto) the keyType supported are KEY_TYPE_ECDSA_P_256, KEY_TYPE_ECDSA_P_521, KEY_TYPE_ED25519,KEY_TYPE_RSA_2048, KEY_TYPE_RSA_4096.

Earlier there was no differentiation on the basis of bit size. This was brought in by @nmahabaleshwar [commit](https://github.com/openconfig/gnsi/commit/eb38162f1f5e5b36159a4f8810d98032f864fad8) for ecdsa. And then another commit by @marcushines [commit](https://github.com/openconfig/gnsi/pull/80/commits) that was for rsa key type for 4096 and 2048.

I believe we are still missing support for RSA 3072 and ECDSA_P_384.

dplore:
Hi Rohit, Feel free to create a PR to add the additional keys.  We can
review and approve

rohit21187:
Thanks Darren

morrowc:
#206 was merged.

PR #206, opened 2024-09-19
Added support for rsa 3072 and ecdsa 384

rohit21187:
The bit size were not supported in the current proto.  #205 
I think we should support these bit size as well

### DIFF
--- a/credentialz/credentialz.proto
+++ b/credentialz/credentialz.proto
@@ -429,6 +429,8 @@ enum KeyType {
   KEY_TYPE_ED25519 = 3;
   KEY_TYPE_RSA_2048 = 4;
   KEY_TYPE_RSA_4096 = 5;
+  KEY_TYPE_RSA_3072 =6;
+  KEY_TYPE_ECDSA_P_384=7;
 }
 
 message AccountCredentials {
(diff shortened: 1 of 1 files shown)


credentialz/credentialz.proto
```diff
@@ -429,6 +429,8 @@ enum KeyType {
   KEY_TYPE_ED25519 = 3;
   KEY_TYPE_RSA_2048 = 4;
   KEY_TYPE_RSA_4096 = 5;
+  KEY_TYPE_RSA_3072 = 6;
```
morrowc, line 432: What's the reason to support 3072? "get all the pokemons" or because of some actual security benefit over 4096? (my recollection is that 3072 is barely considered safe today, and with post-quantum coming to eat our pokemons this is not safe 'shortly'.)

rohit21187: As we already support 2048 so not supporting 3072 because of security consideration doesn’t seem right to me.
With quantum computing we can’t say 4096 will make a difference when compared to 3072. And it’s a proto so we should let users have more control on the keys that they want to use. In addition, 3072 is a FIPS allowed so I think we should have this.

morrowc (approved):

PR #207, opened 2024-10-29
Add credentialz endpoints to manage system role accounts.

avanpo:
It seems there is no way to manage system role accounts via openconfig today, meaning it's necessary to know about the NOS or underlying OS in order to get that done.

### DIFF
--- a/credentialz/credentialz.proto
+++ b/credentialz/credentialz.proto
@@ -224,6 +224,16 @@ service Credentialz {
   // Get the current public keys from the host. Each configured key will be
   // returned in the provided list.
   rpc GetPublicKeys(GetPublicKeysRequest) returns (GetPublicKeysResponse);
+
+  // Add a system role account.
+  // Note that an attempt to add a system role account that already exists will
+  // be rejected with an error.
+  rpc AddAccount(AddAccountRequest) returns (AddAccountResponse);
+
+  // Delete a system role account.
+  // Note that an attempt to delete a system role account that doesn't exist
+  // will be rejected with an error.
+  rpc DeleteAccount(DeleteAccountRequest) returns (DeleteAccountResponse);
 }
 
 //
@@ -787,3 +797,17 @@ message GetPublicKeysResponse {
   // Return all currently configured keys on host.
   repeated PublicKey public_keys = 1;
 }
+
+message AddAccountRequest {
+  // The system role account name. This account must not already exist.
+  string account = 1;
+}
+
+message AddAccountResponse {}
+
+message DeleteAccountRequest {
+  // The system role account name. This account must exist.
+  string account = 1;
+}
+
+message DeleteAccountResponse {}


credentialz/credentialz.proto
```diff
@@ -224,6 +224,16 @@ service Credentialz {
   // Get the current public keys from the host. Each configured key will be
   // returned in the provided list.
   rpc GetPublicKeys(GetPublicKeysRequest) returns (GetPublicKeysResponse);
+
+  // Add a system role account.
```
LimeHat, line 228: Would be good to add a strict definition what a "system role account" is.

OpenConfig has `SYSTEM_ROLE_ADMIN` https://github.com/openconfig/public/blob/db78d44e93a97cef4db27f82bba3fba454e907a5/release/models/system/openconfig-aaa-types.yang#L77-L84, but I have a feeling that you mean something else in this context.

morrowc: Is it the case, avanpo, that you mean here:
 "Add an account to the NOS, but not to the underlying unix system."
 
 This maybe an operator account:
 
    operator - robot/user responsible for interface up/down/troubleshooting

or an administrative account:

   god - the user with all power/privilege on the device

morrowc: ping when you get time.

morrowc (approved):

credentialz/credentialz.proto
```diff
@@ -787,3 +797,17 @@ message GetPublicKeysResponse {
   // Return all currently configured keys on host.
   repeated PublicKey public_keys = 1;
 }
+
+message AddAccountRequest {
+  // The system role account name. This account must not already exist.
+  string account = 1;
```
nmahabaleshwar, line 803: Some implementations need a bare minimum password to create an account, would suggest add `Password` which can be optional.

We might also want to clarify if `Password` is not specified we want to create an account with,
1. Disabled password (prevent authentication via password) OR
2. No password (password not required for authentication)

credentialz/credentialz.proto
LimeHat, line 228: I'm not sure why my previous comment was marked as resolved, but I still would like to get the answer to that question.

Would be good to add a strict definition what a "system role account" is.

OpenConfig has SYSTEM_ROLE_ADMIN https://github.com/openconfig/public/blob/db78d44e93a97cef4db27f82bba3fba454e907a5/release/models/system/openconfig-aaa-types.yang#L77-L84, but I have a feeling that you mean something else in this context.

LimeHat (changes requested):

credentialz/credentialz.proto
morrowc: (picking long past old PR)

Either specifying in the proto/docs:
  "If a passwd is required, but not provided, an error is returned"
or:
   "where a password is required and not provided the device will provide an unusable password"
     (like `username:*:userid:groupid:comment:shell` on standard unix-like platforms)

seems ok.

morrowc (commented):
Left 2 clarifying answers/pointers to the existing conversations.

credentialz/credentialz.proto
LimeHat: >I think the intent of this is to provide a way to set a passwd/ssh-key/etc for:

is it? the proto says it can only add or delete account. for credentials management we have [gnsi.credz](https://github.com/openconfig/gnsi/tree/main/credentialz) service..

morrowc (approved):

credentialz/credentialz.proto
morrowc: I suspect that 'system role account' here might have meant (and I'm sure i mis-read this originally):
  "I want to add a user to the config"
 or:
   "I want to delete a user from the config"
   
  Today credentialz can't really add or remove accounts. This change should permit us to get to that state, right?

morrowc: Ok, back looking at this after some other internal questions.

I think what we really want is:
  "I need to add/delete users from this device"
  
we don't have that today :( at least not in the gnsi tree. 

@LimeHat if we specify that this is 'any account' can be added/deleted, where does that leave you in the questions/answers? (if 'any account', then we need to change the PR descr)

PR #209, opened 2024-11-06
Add API to retrieve the TCG Reference Integrity Manifest data.

avanpo:
Can be used as a prerequisite for certificate provisioning. See https://trustedcomputinggroup.org/resource/tcg-reference-integrity-manifest-rim-information-model/.


marcushines:
you need to regenerate the pb.go files
also can you update the readme to explain the exepected workflow for this new api

avanpo:
Thanks Marcus, I've added a blurb to the readme (and fixed up the API, I was missing a field). PTAL.

certz/certz.proto
```diff
@@ -636,8 +641,84 @@ message CertificateSigningRequest {
...
+  // This is a packed TCG data structure.
+  bytes signature = 2;
+
+  // The EK leaf certificate in PEM format.
+  string ek_cert = 3;
+
+  // The intermediate and signing certificate chain in PEM format.
+  string ek_cert_chain = 4;
+
+  // The AK public key.
```
marcushines, line 718: in what format?

certz/certz.proto
```diff
@@ -636,8 +641,84 @@ message CertificateSigningRequest {
...
+// Fields of type bytes are octets; not text, not base64 encoded.
+message MBMData {
+  bytes digest = 1;
+
+  // The signature over the integrity digest.
+  // This is a packed TCG data structure.
+  bytes signature = 2;
+
+  // The EK leaf certificate in PEM format.
+  string ek_cert = 3;
```
marcushines, line 713: we have had several issues with people "interpreting" the format of these messages

can we make it ek_leaf_cert?

certz/certz.proto
```diff
@@ -636,8 +641,84 @@ message CertificateSigningRequest {
...
+  // Measurement selection structure.
+  // This is a packed TCG data structure.
+  bytes selection = 3;
+
+  // Random nonce used to prevent replay attacks.
+  bytes nonce = 4;
+}
+
+// Measured Boot Manifest measurements.
+// Fields of type bytes are octets; not text, not base64 encoded.
```
marcushines, line 704: can you provide a pointer to those structs? and when you say "octets" are you meaning a byte representation of a c struct?

certz/certz.proto
```diff
@@ -636,8 +641,84 @@ message CertificateSigningRequest {
...
+
+message GetIntegrityManifestRequest {}
+
+message GetIntegrityManifestResponse {
+  bytes manifest = 1;
+}
+
+// See the TCG Reference Integrity Manifest (RIM) Information Model.
+message ReferenceIntegritySpec {
+  // Identifies the Root of Trust component.
```
marcushines, line 666: what are the list of components we consider valid?
should this be a string or a enum

certz/certz.proto
```diff
@@ -636,8 +641,84 @@ message CertificateSigningRequest {
...
+  string rot_id = 1;
+
+  oneof type {
```
marcushines, line 669: what would be the other types we would be considering here?
is this really a oneof?

certz/certz.proto
avanpo: PEM, docstring has been updated.

certz/certz.proto
avanpo: Done.

certz/certz.proto
avanpo: As discussed offline, this should remain a string.

certz/certz.proto
avanpo: Yes, I mean a byte representation of a TSS struct. I've added the relevant links and struct names in the docstrings.

certz/certz.proto
avanpo: I think the idea is that there could be alternatives to MBM, so the oneof leaves that possibility open.

certz/certz.proto
```diff
@@ -636,8 +641,108 @@ message CertificateSigningRequest {
...
+
+  // Reference integrity measurements as specified in the request.
+  repeated ReferenceIntegrityData integrity = 2;
+}
+
+message GetIntegrityManifestRequest {}
+
+message GetIntegrityManifestResponse {
+  // The manifest is a generic byte stream for passing protocol dependent
+  // (unsecured) data required prior to returning attestation data.
```
marcushines, line 662: can we add enough a comment to say this will migrated to structured data that will be a reference once defined

marcushines (approved):

certz/certz.proto
avanpo: Done.

PR #211, opened 2024-11-26
Issue210: Update / Improve comment for 'AccountCredentials'.

morrowc:
Update the docuemntation for AccountCredential to make it clear what the parts of the proto message mean, with respect to a normal openssh format ssh public key.

### DIFF
--- a/credentialz/credentialz.proto
+++ b/credentialz/credentialz.proto
@@ -433,11 +433,11 @@ enum KeyType {
   KEY_TYPE_ECDSA_P_384 = 7;
 }
 
+// AuthorizedKey is an ssh key/idenity. The basic openssh format follows:
+// ecdsa-sha2-nistp521 AAAkeystuffhere chuck@chuckfinley.org
+//   key_type         authorized_key    description
 message AccountCredentials {
   message AuthorizedKey {
-    // AuthorizedKey is an ssh key/idenity. The basic openssh format follows:
-    // ecdsa-sha2-nistp521 AAAkeystuffhere chuck@chuckfinley.org
-    //   key_type         authorized_key    description
     bytes authorized_key = 1;
     // Options specified for this authorized key.
     repeated Option options = 2;
--- a/credentialz/credentialz.proto
+++ b/credentialz/credentialz.proto
@@ -435,6 +435,9 @@ enum KeyType {
 
 message AccountCredentials {
   message AuthorizedKey {
+    // AuthorizedKey is an ssh key/idenity. The basic openssh format follows:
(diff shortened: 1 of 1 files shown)


morrowc:
I'd assign this to @limehat but apparently I can't do that?

morrowc:
I'd also try to assign this to @melzhan but I can't do that either? 

how does review / assignment / review even work here?

dplore:
> I'd also try to assign this to @melzhan but I can't do that either?
> 
> how does review / assignment / review even work here?

I created a team: https://github.com/orgs/openconfig/teams/gnsi-contributors/members
Gave it "triage" permissions (to manage PR's and Issues, but not write access).
I added @melzhan and @LimeHat to this contributors group.  I made @morrowc and admin, so you can add/remove people from here.

There is a maintainers team (https://github.com/orgs/openconfig/teams/gnsi-maintainers) which has write access.

melzhan (approved):

credentialz/credentialz.proto
```diff
@@ -435,6 +435,9 @@ enum KeyType {
 
 message AccountCredentials {
   message AuthorizedKey {
+    // AuthorizedKey is an ssh key/idenity. The basic openssh format follows:
+    // ecdsa-sha2-nistp521 AAAkeystuffhere chuck@chuckfinley.org
+    //   key_type         authorized_key    description
     bytes authorized_key = 1;
```
LimeHat, line 441: The format of this field (`authorized_key`) should be explicitly indicated, as Mel noted above, I think the above comment refers to the message instead

LimeHat (changes requested):
If the non-zero key_type is mandatory, please indicate that in the comment.

In addition, if you want to document the message as a whole, note that the mapping between the enum `key_type` and the openssh key type is N:1 instead of 1:1 (for whatever reason we have three different RSA key types; all of which will be mapped to `ssh-rsa` in the openssh authorized_keys file format.

credentialz/credentialz.proto
```diff
@@ -435,6 +435,9 @@ enum KeyType {
 
 message AccountCredentials {
   message AuthorizedKey {
+    // AuthorizedKey is an ssh key/idenity. The basic openssh format follows:
```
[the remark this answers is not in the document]
morrowc: done.

credentialz/credentialz.proto
morrowc: done, moved the comment up for mel's comment and clarified that the key-type is mandatory.

melzhan:
It looks like I can't approve this and it needs approval from Rob or Marcus

PR #213, opened 2024-12-17
Add support for profiles in Get and Probe RPC.

LarsxGitHub:
While profiles have been added into the rotation via #200, the Get and Probe RPCs have remained profile-less. If a client is able to add multiple profiles, it should also be able to get and probe these profiles individually.


authz/authz.proto
```diff
@@ -230,6 +228,16 @@ message ProbeRequest {
...
+
+  // The profile for which the authz policy is being probed. In the case that
+  // this field is not specified, the default authz policy which applies to all
+  // gRPC endpoints run by the target is assumed. Where non-default policies
+  // are to supported by an endpoint, the value of the profile determines which
+  // set of policies are to be probed.
+  //
+  // Note that the authz profile is considered independent from a SSL profile
+  // ID (as referenced by gnsi.Certz).
+  string authz_profile_id = 3;
```
robshakir, line 240: I discussed with Marcus too. I think this makes sense to have here so that you can know exactly what authz policy you're probing.

However, can we add a clarification that says that you can only have one rotate operation in progress at once, to avoid the complexity of the state machine having to handle "what happens if you start multiple operations and then probe another one".

robshakir: I made a similar change for Certz here: https://github.com/openconfig/gnsi/pull/201

marcushines: +1

authz/authz.proto
```diff
@@ -247,10 +255,20 @@ message ProbeResponse {
...
 message GetRequest {
+  // The profile for which the authz policy is being requested. In the case that
+  // this field is not specified, the default authz policy which applies to all
+  // gRPC endpoints run by the target is assumed. Where non-default policies
+  // are to supported by an endpoint, the value of the profile determines which
+  // set of policies are returned.
+  //
+  // Note that the authz profile is considered independent from a SSL profile
+  // ID (as referenced by gnsi.Certz).
+  string authz_profile_id = 1;
```
morrowc, line 271: Is this really 'ssl_profile_id'? or are there now new profiles inside profiles?
(authz inside ssl)

wait: "Note that the authz profile is considered..."

Why is that? are these tied back to the ssl profile in some way? how would I know that I'm using the a vs b profile here?

morrowc:
This got lost :( does this still need to be done?
(if so I think there's some set of merge issues in the generated bindings, which are likely easier to fix by just re-making the bindings)

morrowc:
Hey there, still working on this? or cancel?

PR #214, opened 2024-12-28
Issue202: Add clarification on usage of ssl profiles.

morrowc:
Add some clarification that SSL Profiles maybe used for:
  * client services (services on a system which initiate connections)
  * server services (services on a system which receive connections)
  * both (clients and servers)
  
This should satisfy Issue202.


marcushines:
can you rebase this PR? that should fix build if still needed

morrowc:
> can you rebase this PR? that should fix build if still needed

Possibly done.

marcushines (approved):

PR #220, opened 2025-01-14
Update build to support Bazel 8/bzlmod

bstoll:
This PR updates GNSI to use Bazel8/bzlmod.

As a part of this change, C++ targets are moved off of rules_proto_grpc and to the cc_proto_library / cc_grpc_library supported by the Protobuf/GRPC packages. This may be a breaking name change for C++ users.

This PR is flagged as a draft because it pins openconfig_gnmi to my personal git branch. openconfig/gnmi#192 needs to be merged and uploaded to the Bazel Central Registry before this PR can be merged.


morrowc (approved):
sure... bazel things are bazel things...

I assume this was done with some magical script and not by typing like a mad persona?

bstoll:
Gazelle can generate most of the BUILD files (`bazel run //:gazelle`) and `gnsi_go_deps.bzl` is generated from go.mod (`bazel mod tidy` & `bazel run //:gazelle_update_repos`). The remaining is mostly not generated.

I think the biggest change here is the change to C++ deps: if a build previously depended on something like `//acctz:acctz_cc_proto` and uses GRPC, it will probably need to add //acctz:acctz_cc_grpc_proto to the deps list.  This is due to switching from using the community-maintained [rules_proto_grpc](https://github.com/rules-proto-grpc/rules_proto_grpc) to using the rules from the main protocolbuffers/GRPC releases.

PR #247, opened 2025-09-16
Fix timestamp abiguity for all services

marcushines:
the OC yang models these as nanos
The tests also test these created_on fields as nanos we need to update the docstrings to align with this design.	
Test:
   CreatedOn: uint64(time.Now().Unix()),

This needs to be universally one or the other and some tests are using nanos


robshakir (approved):
Given that it's already nanoseconds in the tests, this just looks to be housekeeping and doesn't result in any incompatibilities. Approving.

PR #260, opened 2026-02-13
Add enums for self-signed cert+key generated on device.

Chounoki:
The new enums are for cases where we need the device to generate a self-signed cert+key pair directly on the device when conveying a self-signed cert+key over an insecure connection is risky.


morrowc (approved):

Chounoki:
Thanks for reviewing!

Issue #266, opened 2026-03-24
Clarify SSL Profile content uniqueness/non-uniqueness.


morrowc:
Add some documentation to clarify that the content of a particular SSL Profile as defined and deployed to a device are not meant to be either unique (unique cert / key / bundle / etc) nor non-repeated among other SSL Profiles deployed to that device.

It may be the case that an operator wants to define a single SSL Profile for all services to use for a device configuration, in that case there may be unique tls artifacts deployed. It may also be the case that an operator may choose to deploy a unique SSL Profile per TLS enabled service on a device. In that case there MAY be unique artifacts, or there may be sharing / duplication of those artifacts across the various SSL Profiles configured on the device.

PR #267, opened 2026-03-24
Issue #266: Documentation Update: SSL Profile clarity.

morrowc:
Clarify that the content in an SSL Profile is in no way intended to be unique or non-repeating among the various SSL Profiles configured on a particular system.

See the content in Issue 266.


morrowc (commented):
I believe this is ready for review.

PR #271, opened 2026-04-30
Add profile ID in Get Request

goabhinav:
We need this field to implement Get() by profile ID for the INT test of dsdn authz.


alshabib (approved):

authz/authz.proto
```diff
@@ -251,6 +251,8 @@ message ProbeResponse {
 
 // GetRequest used to request the gRPC-level Authorization Policy.
 message GetRequest {
+  // The profile for which the authz policy is being requested.
+  string authz_profile_id = 1;
```
morrowc, line 255: should we return this in the `response,` too?

morrowc (approved):

authz/authz.proto
goabhinav: At the moment we don't need this in response! Do you think it will be a useful addition in some way? Otherwise can we merge this?

morrowc: I suppose I was thinking that when getting a reply you might want to know: "Oh my request was for ssl profileX and the reply was also for that"

Since the requests / replies here are part of a stream it's not necessarily guaranteed that the request/reply are linked in time, correct?

goabhinav: I see the Get rpc is unary and not a streaming rpc. So I think the request and reply are inherently linked in time. The reply should be a direct response to the specific request that was sent. Please correct me if my understanding is wrong!

morrowc: oh! sure this isn't part of the rotate biz.. cool!

morrowc (approved):

PR #275, opened 2026-07-22
Remove comment about x509 CRL type - they are all x509, afaik, since

haussli:
these protos only handle SSL/TLS certificates, which are all x509.


mihirpitale-googler (approved):
lgtm

